PR #55, opened 2020-01-31
fix: refactor: Fix issue with storeVC and refactor BDD tests

DRK3:
Fixed an issue that was causing VCs to not be stored properly.

Refactored BDD tests.


test/bdd/pkg/vc/vc_steps.go
```diff
@@ -100,16 +109,28 @@ func (e *Steps) getProfile(profileID, expectedProfileResponseArgkey string) erro
...
 }
 
-func (e *Steps) createCredential(credentialRequestArgKey, expectedIssuerID, expectedIssuerName string) error {
-	credentialRequest := e.bddContext.Args[credentialRequestArgKey]
+func (e *Steps) createCredential(profileName string) error {
+	credentialRequest := operation.CreateCredentialRequest{}
+
+	err := json.Unmarshal(e.bddContext.CreateCredentialRequestTemplate, &credentialRequest)
+	if err != nil {
+		return nil
```
sandrask: return error

test/bdd/pkg/vc/vc_steps.go
```diff
@@ -125,31 +146,28 @@ func (e *Steps) createCredential(credentialRequestArgKey, expectedIssuerID, expe
...
 
-	if vc.Issuer.Name != expectedIssuerName {
-		return fmt.Errorf("expected %s but got %s instead", expectedIssuerName, vc.Issuer.Name)
-	}
+	storeRequest.Profile = profileName
+	storeRequest.Credential = string(e.bddContext.CreatedCredential)
 
-	return nil
-}
+	storeRequest.Profile = profileName
```
sandrask: duplicate code

test/bdd/pkg/vc/vc_steps.go
```diff
@@ -163,8 +181,18 @@ func (e *Steps) storeCredential(storeVCRequestArgKey string) error {
...
 
-func (e *Steps) retrieveCredential(credentialID, profileName, expectedIssuerID, expectedIssuerName string) error {
-	escapedCredentialID := url.PathEscape(credentialID)
+func (e *Steps) retrieveCredential(profileName string) error {
+	vcMap, err := getVCMap(e.bddContext.CreatedCredential)
+	if err != nil {
+		return err
+	}
+
+	vcID, found := vcMap["id"].(string)
```
sandrask: this also covers the case when id is present but of wrong type so error message is bit misleading

test/bdd/pkg/vc/vc_steps.go
```diff
@@ -242,41 +263,85 @@ func (e *Steps) verifyCredential(validVCArgKey string) error {
...
 
-func (e *Steps) checkProfileResponse(expectedProfileResponseArgkey string,
+func (e *Steps) checkProfileResponse(expectedProfileResponseName string,
 	profileResponse *operation.ProfileResponse) error {
-	expectedProfileResponse := operation.ProfileResponse{}
+	if profileResponse.Name != expectedProfileResponseName {
+		return fmt.Errorf("expected %s but got %s instead", expectedProfileResponseName, profileResponse.Name)
+	}
+
+	expectedProfileResponseDID := "did:peer:22"
```
sandrask: it should be constant, same applies bellow

test/bdd/pkg/vc/vc_steps.go
DRK3: oops. Fixed!

test/bdd/pkg/vc/vc_steps.go
DRK3: oops. Fixed!

test/bdd/pkg/vc/vc_steps.go
DRK3: You're right. I separated the map get and the type assertion into separate lines so I can check both errors separately. I also refactored the other type assertion I do in that file so that I can check the two errors separately.

test/bdd/pkg/vc/vc_steps.go
DRK3: done

DRK3: done. I also split up the `checkVC` function.

PR #71, opened 2020-02-07
feat: Use base58 encoded 128-bit values for document IDs

### LINKED ISSUE #66: Use base58-encoded 128-bit value when creating documents in EDV
When https://github.com/trustbloc/edv/issues/28 is done, edge-service will need to be updated accordingly.

DRK3:
closes #66

VC-service now encodes document IDs to conform with the new version of the EDV server.

Updated the BDD tests to use the latest EDV image.

Fixed some incorrect error messages in the BDD-tests.

Fixed a few places in the retrieve document handler where errors occurred and the function didn't return after writing the error response.

### DIFF
--- a/pkg/restapi/vc/operation/operations.go
+++ b/pkg/restapi/vc/operation/operations.go
@@ -18,8 +18,10 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
+	"github.com/btcsuite/btcutil/base58"
 	"github.com/google/uuid"
 	"github.com/gorilla/mux"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/ed25519signature2018"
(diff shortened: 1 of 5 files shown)


pkg/restapi/vc/operation/operations.go
```diff
@@ -270,6 +284,22 @@ func (o *Operation) storeVCHandler(rw http.ResponseWriter, req *http.Request) {
...
 
+// This function expects the vcID that's passed in to be a URL with a "/" and then a UUID concatenated to the end.
+func convertVCIDToEDVCompatibleFormat(vcID string) (string, error) {
+	vcIDParts := strings.Split(vcID, "/")
+
+	uuidFromVCID := vcIDParts[len(vcIDParts)-1]
+
+	parsedUUID, err := uuid.Parse(uuidFromVCID)
+	if err != nil {
+		return "", fmt.Errorf(fmt.Sprintf("%s: %s", invalidUUIDErrMsg, err.Error()))
```
Baha-sk: fmt.Sprintf is redundant here... Errorf() does the same thing

DRK3: Oops, good catch. Fixed.

PR #92, opened 2020-02-28
fix: upload code cov

fqutishat:

### DIFF
--- a/ci/azure-pipelines.yml
+++ b/ci/azure-pipelines.yml
@@ -28,7 +28,9 @@ jobs:
       - checkout: self
       - script: make unit-test
         displayName: Run unit test
-      - script: bash <(curl https://codecov.io/bash)
+      - script: bash <(curl https://codecov.io/bash) -t $CODECOV_UPLOAD_TOKEN
+        env:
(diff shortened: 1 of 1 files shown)


ci/azure-pipelines.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - checkout: self
       - script: make unit-test
         displayName: Run unit test
-      - script: bash <(curl https://codecov.io/bash)
+      - script: bash <(curl https://codecov.io/bash) -t 276dd96d-96a9-4518-b7f7-5028a2ddf658
```
llorllale: can we keep this a secret in the pipeline?

DRK3: I actually remember reading somewhere that it's possible to do that...but I'm having trouble finding the link...

DRK3: Although according to the documentation, Azure Pipelines shouldn't even need a token. It's weird that it doesn't work without it anymore.

https://docs.codecov.io/docs/about-the-codecov-bash-uploader

fqutishat: done

PR #102, opened 2020-03-06
feat: Encrypt documents sent to EDV

### LINKED ISSUE #61: Encrypt documents sent to EDV
Using Aries legacy JWE for now.

DRK3:
- Many of the operations unit tests were using a mocked KMS when they could have used a real one - those have been updated to use real in-memory KMSs now.
- Added shorthand flags for the Bloc domain, external URL host, and universal resolver URL flags.
- Some other minor refactoring and fixes

closes #61

### DIFF
--- a/cmd/vc-rest/startcmd/start.go
+++ b/cmd/vc-rest/startcmd/start.go
@@ -31,28 +31,36 @@ import (
 )
 
 const (
-	hostURLFlagName          = "host-url"
-	hostURLFlagShorthand     = "u"
-	hostURLFlagUsage         = "URL to run the vc-rest instance on. Format: HostName:Port."
-	hostURLEnvKey            = "VC_REST_HOST_URL"
-	edvURLFlagName           = "edv-url"
-	edvURLFlagShorthand      = "e"
-	edvURLFlagUsage          = "URL EDV instance is running on. Format: HostName:Port."
-	edvURLEnvKey             = "EDV_REST_HOST_URL"
-	blocDomainFlagName       = "bloc-domain"
-	blocDomainFlagUsage      = "Bloc domain"
-	blocDomainEnvKey         = "BLOC_DOMAIN"
-	hostURLExternalFlagName  = "host-url-external"
-	hostURLExternalEnvKey    = "VC_REST_HOST_URL_EXTERNAL"
-	hostURLExternalFlagUsage = "Host External Name:Port This is the URL for the host server as seen externally." +
+	hostURLFlagName      = "host-url"
+	hostURLFlagShorthand = "u"
+	hostURLFlagUsage     = "URL to run the vc-rest instance on. Format: HostName:Port."
+	hostURLEnvKey        = "VC_REST_HOST_URL"
+
+	edvURLFlagName      = "edv-url"
+	edvURLFlagShorthand = "e"
+	edvURLFlagUsage     = "URL EDV instance is running on. Format: HostName:Port."
+	edvURLEnvKey        = "EDV_REST_HOST_URL"
+
+	blocDomainFlagName      = "bloc-domain"
+	blocDomainFlagShorthand = "b"
+	blocDomainFlagUsage     = "Bloc domain"
+	blocDomainEnvKey        = "BLOC_DOMAIN"
+
+	hostURLExternalFlagName      = "host-url-external"
+	hostURLExternalFlagShorthand = "x"
+	hostURLExternalEnvKey        = "VC_REST_HOST_URL_EXTERNAL"
+	hostURLExternalFlagUsage     = "Host External Name:Port This is the URL for the host server as seen externally." +
 		" If not provided, then the host url will be used here." +
 		" Alternatively, this can be set with the following environment variable: " + hostURLExternalEnvKey
-	universalResolverURLFlagName  = "universal-resolver-url"
-	universalResolverURLFlagUsage = "Universal Resolver instance is running on. Format: HostName:Port."
-	universalResolverURLEnvKey    = "UNIVERSAL_RESOLVER_HOST_URL"
-	modeFlagName                  = "mode"
-	modeFlagShorthand             = "m"
-	modeFlagUsage                 = "Mode in which the vc-rest service will run. Possible values: " +
+
+	universalResolverURLFlagName      = "universal-resolver-url"
+	universalResolverURLFlagShorthand = "r"
+	universalResolverURLFlagUsage     = "Universal Resolver instance is running on. Format: HostName:Port."
+	universalResolverURLEnvKey        = "UNIVERSAL_RESOLVER_HOST_URL"
+
+	modeFlagName      = "mode"
+	modeFlagShorthand = "m"
+	modeFlagUsage     = "Mode in which the vc-rest service will run. Possible values: " +
 		"['issuer', 'verifier'] (default: issuer)."
 	modeEnvKey = "VC_REST_MODE"
 
@@ -68,7 +76,6 @@ const (
 )
 
 type vcRestParameters struct {
-	srv                  server
 	hostURL              string
 	edvURL               string
 	blocDomain           string
@@ -104,70 +111,78 @@ func createStartCmd(srv server) *cobra.Command {
 		Short: "Start vc-rest",
 		Long:  "Start vc-rest inside the edge-service",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			hostURL, err := cmdutils.GetUserSetVar(cmd, hostURLFlagName, hostURLEnvKey, false)
+			parameters, err := getVCRestParameters(cmd)
 			if err != nil {
 				return err
 			}
 
-			edvURL, err := cmdutils.GetUserSetVar(cmd, edvURLFlagName, edvURLEnvKey, false)
-			if err != nil {
-				return err
-			}
-
-			blocDomain, err := cmdutils.GetUserSetVar(cmd, blocDomainFlagName, blocDomainEnvKey, false)
-			if err != nil {
-				return err
-			}
-
-			hostURLExternal, err := cmdutils.GetUserSetVar(cmd, hostURLExternalFlagName,
-				hostURLExternalEnvKey, true)
-			if err != nil {
-				return err
-			}
-
-			universalResolverURL, err := cmdutils.GetUserSetVar(cmd, universalResolverURLFlagName,
-				universalResolverURLEnvKey, true)
-			if err != nil {
-				return err
-			}
-
-			mode, err := cmdutils.GetUserSetVar(cmd, modeFlagName, modeEnvKey, true)
-			if err != nil {
-				return err
-			}
-
-			if !supportedMode(mode) {
-				return fmt.Errorf("unsupported mode: %s", mode)
-			}
-
-			if mode == "" {
-				mode = string(issuer)
-			}
-
-			parameters := &vcRestParameters{
-				srv:                  srv,
-				hostURL:              hostURL,
-				edvURL:               edvURL,
-				blocDomain:           blocDomain,
-				hostURLExternal:      hostURLExternal,
-				universalResolverURL: universalResolverURL,
-				mode:                 mode,
-			}
-			return startEdgeService(parameters)
+			return startEdgeService(parameters, srv)
 		},
 	}
+}
+
+func getVCRestParameters(cmd *cobra.Command) (*vcRestParameters, error) {
+	hostURL, err := cmdutils.GetUserSetVar(cmd, hostURLFlagName, hostURLEnvKey, false)
+	if err != nil {
+		return nil, err
+	}
+
+	edvURL, err := cmdutils.GetUserSetVar(cmd, edvURLFlagName, edvURLEnvKey, false)
+	if err != nil {
+		return nil, err
+	}
+
+	blocDomain, err := cmdutils.GetUserSetVar(cmd, blocDomainFlagName, blocDomainEnvKey, false)
+	if err != nil {
+		return nil, err
+	}
+
+	hostURLExternal, err := cmdutils.GetUserSetVar(cmd, hostURLExternalFlagName,
+		hostURLExternalEnvKey, true)
+	if err != nil {
+		return nil, err
+	}
+
+	universalResolverURL, err := cmdutils.GetUserSetVar(cmd, universalResolverURLFlagName,
+		universalResolverURLEnvKey, true)
+	if err != nil {
+		return nil, err
+	}
+
+	mode, err := cmdutils.GetUserSetVar(cmd, modeFlagName, modeEnvKey, true)
+	if err != nil {
+		return nil, err
+	}
+
+	if !supportedMode(mode) {
+		return nil, fmt.Errorf("unsupported mode: %s", mode)
+	}
+
+	if mode == "" {
+		mode = string(issuer)
+	}
+
+	return &vcRestParameters{
+		hostURL:              hostURL,
+		edvURL:               edvURL,
+		blocDomain:           blocDomain,
+		hostURLExternal:      hostURLExternal,
+		universalResolverURL: universalResolverURL,
+		mode:                 mode,
+	}, nil
 }
 
 func createFlags(startCmd *cobra.Command) {
 	startCmd.Flags().StringP(hostURLFlagName, hostURLFlagShorthand, "", hostURLFlagUsage)
 	startCmd.Flags().StringP(edvURLFlagName, edvURLFlagShorthand, "", edvURLFlagUsage)
-	startCmd.Flags().StringP(blocDomainFlagName, "", "", blocDomainFlagUsage)
-	startCmd.Flags().StringP(hostURLExternalFlagName, "", "", hostURLExternalFlagUsage)
-	startCmd.Flags().StringP(universalResolverURLFlagName, "", "", universalResolverURLFlagUsage)
+	startCmd.Flags().StringP(blocDomainFlagName, blocDomainFlagShorthand, "", blocDomainFlagUsage)
+	startCmd.Flags().StringP(hostURLExternalFlagName, hostURLExternalFlagShorthand, "", hostURLExternalFlagUsage)
+	startCmd.Flags().StringP(universalResolverURLFlagName, universalResolverURLFlagShorthand, "",
+		universalResolverURLFlagUsage)
 	startCmd.Flags().StringP(modeFlagName, modeFlagShorthand, "", modeFlagUsage)
 }
 
-func startEdgeService(parameters *vcRestParameters) error {
+func startEdgeService(parameters *vcRestParameters, srv server) error {
 	// Create KMS
 	kms, err := createKMS(mem.NewProvider())
 	if err != nil {
@@ -201,7 +216,7 @@ func startEdgeService(parameters *vcRestParameters) error {
 
 	log.Infof("Starting vc rest server on host %s", parameters.hostURL)
 
-	return parameters.srv.ListenAndServe(parameters.hostURL, router)
+	return srv.ListenAndServe(parameters.hostURL, router)
 }
 
 func createKMS(s storage.Provider) (ariesapi.CloseableKMS, error) {
--- a/cmd/vc-rest/startcmd/start_test.go
+++ b/cmd/vc-rest/startcmd/start_test.go
@@ -93,7 +93,6 @@ func TestStartCmdWithMissingArg(t *testing.T) {
 			"Neither host-url (command line flag) nor VC_REST_HOST_URL (environment variable) have been set.",
 			err.Error())
 	})
-
 	t.Run("test missing edv url arg", func(t *testing.T) {
 		startCmd := GetStartCmd(&mockServer{})
 
@@ -151,31 +150,12 @@ func TestStartCmdValidArgs(t *testing.T) {
 func TestStartCmdValidArgsEnvVar(t *testing.T) {
 	startCmd := GetStartCmd(&mockServer{})
 
-	err := os.Setenv(hostURLEnvKey, "localhost:8080")
-	require.Nil(t, err)
+	setEnvVars(t)
 
-	err = os.Setenv(edvURLEnvKey, "localhost:8081")
-	require.Nil(t, err)
+	defer unsetEnvVars(t)
 
-	err = os.Setenv(blocDomainEnvKey, "domain")
-	require.Nil(t, err)
-
-	err = startCmd.Execute()
-
-	require.Nil(t, err)
-}
-
-func checkFlagPropertiesCorrect(t *testing.T, cmd *cobra.Command, flagName, flagShorthand, flagUsage string) {
-	flag := cmd.Flag(flagName)
-
-	require.NotNil(t, flag)
-	require.Equal(t, flagName, flag.Name)
-	require.Equal(t, flagShorthand, flag.Shorthand)
-	require.Equal(t, flagUsage, flag.Usage)
-	require.Equal(t, "", flag.Value.String())
-
-	flagAnnotations := flag.Annotations
-	require.Nil(t, flagAnnotations)
+	err := startCmd.Execute()
+	require.NoError(t, err)
 }
 
 func TestCreateVDRI(t *testing.T) {
@@ -187,7 +167,7 @@ func TestCreateVDRI(t *testing.T) {
 	})
 
 	t.Run("test error from create new universal resolver vdri", func(t *testing.T) {
-		err := startEdgeService(&vcRestParameters{universalResolverURL: "wrong"})
+		err := startEdgeService(&vcRestParameters{universalResolverURL: "wrong"}, nil)
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "failed to create new universal resolver vdri")
 	})
@@ -213,6 +193,41 @@ func TestCreateKMS(t *testing.T) {
 		require.NoError(t, err)
 		require.NotNil(t, v)
 	})
+}
+
+func setEnvVars(t *testing.T) {
+	err := os.Setenv(hostURLEnvKey, "localhost:8080")
+	require.NoError(t, err)
+
+	err = os.Setenv(edvURLEnvKey, "localhost:8081")
+	require.NoError(t, err)
+
+	err = os.Setenv(blocDomainEnvKey, "domain")
+	require.NoError(t, err)
+}
+
+func unsetEnvVars(t *testing.T) {
+	err := os.Unsetenv(hostURLEnvKey)
+	require.NoError(t, err)
+
+	err = os.Unsetenv(edvURLEnvKey)
+	require.NoError(t, err)
+
+	err = os.Unsetenv(blocDomainEnvKey)
+	require.NoError(t, err)
+}
+
+func checkFlagPropertiesCorrect(t *testing.T, cmd *cobra.Command, flagName, flagShorthand, flagUsage string) {
+	flag := cmd.Flag(flagName)
+
+	require.NotNil(t, flag)
+	require.Equal(t, flagName, flag.Name)
+	require.Equal(t, flagShorthand, flag.Shorthand)
+	require.Equal(t, flagUsage, flag.Usage)
+	require.Equal(t, "", flag.Value.String())
+
+	flagAnnotations := flag.Annotations
+	require.Nil(t, flagAnnotations)
 }
 
 // MockStoreProvider mock store provider.
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,6 @@ require (
 	github.com/spf13/cobra v0.0.6
 	github.com/stretchr/testify v1.4.0
 	github.com/trustbloc/edge-core v0.1.2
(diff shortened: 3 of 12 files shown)


cmd/vc-rest/startcmd/start.go
```diff
@@ -125,7 +125,7 @@ func createStartCmd(srv server) *cobra.Command {
 			}
 
 			universalResolverURL, err := cmdutils.GetUserSetVar(cmd, universalResolverURLFlagName,
-				universalResolverURLEnvKey, true)
+				universalResolverURLEnvKey, false)
```
DRK3: I noticed that this was set to optional, but it seems to be required since I couldn't go through the REST flow without having a universal resolver running, so I set it to required (isOptional=false). @fqutishat  @sandrask Let me know if you have any concerns.

go.mod
```diff
@@ -15,6 +15,5 @@ require (
 	github.com/spf13/cobra v0.0.6
 	github.com/stretchr/testify v1.4.0
 	github.com/trustbloc/edge-core v0.1.2
-	github.com/trustbloc/edv v0.1.2
-	github.com/trustbloc/sidetree-core-go v0.1.2-0.20200226223710-852e23076586
```
DRK3: I ran a `go mod tidy` and it removed that sidetree-core-go line. I guess it was left over from a previous commit.

pkg/restapi/vc/operation/operations_test.go
```diff
@@ -37,14 +43,13 @@ import (
 	"github.com/trustbloc/edge-service/pkg/internal/mock/edv"
 )
 
-const multipleContext = `"@context":["https://www.w3.org/2018/credentials/v1","https://w3id.org/citizenship/v1"]`
-
-const validContext = `"@context":["https://www.w3.org/2018/credentials/v1"]`
-
-const invalidContext = `"@context":"https://www.w3.org/2018/credentials/v1"`
+const (
```
DRK3: I took all the different `const` declarations and grouped them all, so that's why you'll see so many line changes around here.

pkg/restapi/vc/operation/operations_test.go
```diff
@@ -195,9 +184,60 @@ const validVC = `{` +
...
+		validContext + `,
+  "id": "http://example.edu/credentials/1872",
+  "type": "VerifiableCredential",
+  "credentialSubject": {
+    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21"
+  },
+  "issuanceDate": "2010-01-01T19:23:24Z"
+}`
 
-//nolint:lll
```
DRK3: I noticed there were some super long lines in some of these consts, so I split them up into multiple lines. This way we don't need that `nolint`

pkg/restapi/vc/operation/operations_test.go
```diff
@@ -279,61 +321,12 @@ const inValidVCinVP = `{` +
...
-// VC without issuer
-const invalidVC = `{` +
-	validContext + `,
-  "id": "http://example.edu/credentials/1872",
-  "type": "VerifiableCredential",
-  "credentialSubject": {
-    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21"
-  },
-  "issuanceDate": "2010-01-01T19:23:24Z"
-}`
```
DRK3: I moved these up so they would be nearby other related `const`s

pkg/restapi/vc/operation/operations_test.go
```diff
@@ -345,7 +338,7 @@ func TestNew(t *testing.T) {
 	t.Run("test error from open store", func(t *testing.T) {
 		client := edv.NewMockEDVClient("test", nil)
 		op, err := New(&mockstore.Provider{ErrOpenStoreHandle: fmt.Errorf("error open store")},
-			client, &kmsmock.CloseableKMS{}, &vdrimock.MockVDRIRegistry{}, "localhost:8080")
+			client, getTestKMS(t), &vdrimock.MockVDRIRegistry{}, "localhost:8080")
```
DRK3: I noticed that we were using a mock KMS in a lot of places that didn't need a mock - using a real one works fine. Hence you'll see a bunch of these mock KMS calls replaced with a real one.

pkg/restapi/vc/operation/operations_test.go
```diff
@@ -1089,12 +1081,12 @@ func TestStoreVCHandler(t *testing.T) {
 		rr := httptest.NewRecorder()
 		op.storeVCHandler(rr, req)
 		require.Equal(t, http.StatusBadRequest, rr.Code)
-		require.Equal(t, rr.Body.String(), errVaultNotFound.Error())
+		require.Equal(t, rr.Body.String(), "key is nil")
 	})
-	t.Run("store vc err vault not found", func(t *testing.T) {
```
DRK3: This test was a duplicate, so I removed it (the diff makes it look like I "replaced" it with another test, but in reality I removed the duplicate and added a test).

test/bdd/pkg/vc/vc_steps.go
```diff
@@ -329,14 +329,14 @@ func (e *Steps) retrieveCredential(profileName string) error {
...
-	unescapedResponse, err := strconv.Unquote(string(respBytes))
-	if err != nil {
-		return err
-	}
-
 	expectedCredential := string(e.bddContext.CreatedCredential)
-	if unescapedResponse != expectedCredential {
-		return expectedStringError(expectedCredential, unescapedResponse)
+	receivedCredential := string(respBytes)
+	// For some reason there's an extra line at the end of the credential returned from the create credential handler.
```
DRK3: any idea why this is? Normally it doesn't matter since we're just using the JSON unmarshal functions, but it's still odd.

cmd/vc-rest/startcmd/start.go
fqutishat: keep it optional

test/bdd/pkg/vc/vc_steps.go
```diff
@@ -329,14 +329,14 @@ func (e *Steps) retrieveCredential(profileName string) error {
...
+	// By adding a space to the received credential we just received we should be able to directly compare the strings.
+	receivedCredential += `
```
fqutishat: ??

cmd/vc-rest/startcmd/start_test.go
```diff
@@ -160,6 +162,9 @@ func TestStartCmdValidArgsEnvVar(t *testing.T) {
 	err = os.Setenv(blocDomainEnvKey, "domain")
 	require.Nil(t, err)
 
+	err = os.Setenv(universalResolverURLEnvKey, "localhost:8083")
```
Baha-sk: it's a good idea to defer the removal of these test environment variables

test/bdd/pkg/vc/vc_steps.go
Baha-sk: Instead of appending a breakline, compare against trimmed expectedCredential, but definitely this needs to be investigated.

pkg/restapi/vc/operation/operations.go
```diff
@@ -322,29 +346,65 @@ func (o *Operation) storeVCHandler(rw http.ResponseWriter, req *http.Request) {
...
 
 		return
 	}
 }
 
+func (o *Operation) buildStructuredDoc(data *StoreVCRequest,
+	vc *verifiable.Credential) (*operation.StructuredDocument, error) {
+	// We need an EDV-compliant ID. The UUID that's in the VC is a good fit, since it's 128 bits long already.
+	// Let's convert it to base58 and use it as the StructuredDocument ID.
+	edvDocID, err := convertVCIDToEDVCompatibleFormat(vc.ID)
```
troyronda: I’m concerned about this. Is this identifier stored encrypted or unencrypted?

DRK3: This identifier is not encrypted. Let me know your thoughts on this.

test/bdd/pkg/vc/vc_steps.go
DRK3: Done!

cmd/vc-rest/startcmd/start_test.go
DRK3: Done!

cmd/vc-rest/startcmd/start.go
DRK3: I can't reproduce that original issue I was having, so I'll make it optional again

DRK3: Done

test/bdd/pkg/vc/vc_steps.go
DRK3: @fqutishat To clarify: the extra line that I noticed is unrelated to my change, so I created a followup story to investigate and fix it in order to avoid scope creep here.

https://github.com/trustbloc/edge-service/issues/104

Note however that the extra line may be coming from Aries-Framework-Go, since we call into that code while creating a credential. So I (or someone else) might have to make a fix there first.

In the meantime, I've replaced this with the JSON trimming suggestion per @Baha-sk.

cmd/vc-rest/startcmd/start_test.go
```diff
@@ -215,6 +196,41 @@ func TestCreateKMS(t *testing.T) {
 	})
 }
 
+func setEnvVars(t *testing.T) {
+	err := os.Setenv(hostURLEnvKey, "localhost:8080")
```
Baha-sk: also backup the original env value and reset it in unsetEnvVars() if it exists

DRK3: Talked to @Baha-sk, the env vars seem to get wiped after tests anyway, so this isn't needed

pkg/restapi/vc/operation/operations.go
troyronda: Cannot use the credential ID in that case.

DRK3: @sandrask proposed a potential solution. We can hash the VC id and use that instead. This way, you can't derive the VC ID from the EDV document ID. However, if you know the VC ID then you can still retrieve the document. And this way I don't have to store a mapping of VC IDs to EDV document IDs somewhere. 

I've updated the PR with hashing. Let me know what you think. @troyronda @sandrask @Baha-sk

DRK3: (scroll down to the convertVCIDToEDVCompatibleFormat function)

cmd/vc-rest/startcmd/start.go
```diff
@@ -30,31 +30,44 @@ import (
...
-	edvURLFlagShorthand      = "e"
-	edvURLFlagUsage          = "URL EDV instance is running on. Format: HostName:Port."
-	edvURLEnvKey             = "EDV_REST_HOST_URL"
-	blocDomainFlagName       = "bloc-domain"
-	blocDomainFlagUsage      = "Bloc domain"
-	blocDomainEnvKey         = "BLOC_DOMAIN"
-	hostURLExternalFlagName  = "host-url-external"
-	hostURLExternalEnvKey    = "VC_REST_HOST_URL_EXTERNAL"
-	hostURLExternalFlagUsage = "Host External Name:Port This is the URL for the host server as seen externally." +
+	hostURLFlagName      = "host-url"
```
DRK3: A few of these flags were missing their shorthand equivalents. I added them.

I also noticed that there was no spacing between any of these, so I grouped them together based on their purpose.

pkg/internal/mock/edv/client.go
```diff
@@ -12,12 +12,12 @@ import (
...
 type Client struct {
 	edvServerURL            string
-	readDocumentReturnValue []byte
+	ReadDocumentReturnValue *operation.EncryptedDocument
 }
 
 // NewMockEDVClient is the mock version of edv client
-func NewMockEDVClient(edvServerURL string, readDocumentReturnValue []byte) *Client {
-	return &Client{edvServerURL: edvServerURL, readDocumentReturnValue: readDocumentReturnValue}
+func NewMockEDVClient(edvServerURL string, readDocumentReturnValue *operation.EncryptedDocument) *Client {
```
DRK3: This is due to a change in the interface defined in the EDV dependency (which was updated in this PR).

cmd/vc-rest/startcmd/start.go
```diff
@@ -103,46 +117,12 @@ func createStartCmd(srv server) *cobra.Command {
...
-			}
-
-			universalResolverURL, err := cmdutils.GetUserSetVar(cmd, universalResolverURLFlagName,
-				universalResolverURLEnvKey, true)
-			if err != nil {
-				return err
-			}
-
-			mode, err := cmdutils.GetUserSetVar(cmd, modeFlagName, modeEnvKey, true)
+			hostURL, edvURL, blocDomain, hostURLExternal, universalResolverURL, mode, localStorePath, err :=
```
Baha-sk: consider creating a struct with the returned fields and return this struct instead

cmd/vc-rest/startcmd/start.go
```diff
@@ -151,19 +131,72 @@ func createStartCmd(srv server) *cobra.Command {
...
 				mode:                 mode,
+				localStorePath:       localStorePath,
 			}
+
 			return startEdgeService(parameters)
 		},
 	}
 }
 
+func getVCRestParameters(cmd *cobra.Command) (string, string, string, string, string, string, string, error) {
```
Baha-sk: returning more than 2/3 values is a sign they should be grouped in a struct instead.

pkg/restapi/vc/operation/operations.go
```diff
@@ -73,19 +80,27 @@ type vcStatusManager interface {
...
 type Client interface {
 	CreateDataVault(config *operation.DataVaultConfiguration) (string, error)
-	CreateDocument(vaultID string, document *operation.StructuredDocument) (string, error)
-	ReadDocument(vaultID, docID string) ([]byte, error)
+	CreateDocument(vaultID string, document *operation.EncryptedDocument) (string, error)
+	ReadDocument(vaultID, docID string) (*operation.EncryptedDocument, error)
+}
+
+type cryptoProvider struct {
+	crypto legacykms.KeyManager
```
Baha-sk: do not confuse crypto with kms.. they are two different concepts.. KeyManager is a kms, this should be a kmsProvider with a kms field instead of cryptoProvider with crypto field

pkg/restapi/vc/operation/operations.go
```diff
@@ -97,12 +112,33 @@ func New(provider storage.Provider, client Client, kms legacykms.KMS, vdri vdria
 		return nil, fmt.Errorf("failed to instantiate new csl status: %w", err)
 	}
 
+	cryptoProv := cryptoProvider{
+		crypto: kms,
```
Baha-sk: it is more clear when used here... a kms is not a crypto instance

cmd/vc-rest/startcmd/start.go
```diff
@@ -30,31 +30,44 @@ import (
...
+	universalResolverURLFlagUsage     = "Universal Resolver instance is running on. Format: HostName:Port."
+	universalResolverURLEnvKey        = "UNIVERSAL_RESOLVER_HOST_URL"
+
+	modeFlagName      = "mode"
+	modeFlagShorthand = "m"
+	modeFlagUsage     = "Mode in which the vc-rest service will run. Possible values: " +
 		"['issuer', 'verifier'] (default: issuer)."
 	modeEnvKey = "VC_REST_MODE"
 
+	localStorePathFlagName      = "local-store-path"
```
fqutishat: what this flag mean?

pkg/restapi/vc/operation/operations.go
```diff
@@ -322,41 +361,90 @@ func (o *Operation) storeVCHandler(rw http.ResponseWriter, req *http.Request) {
...
-func convertVCIDToEDVCompatibleFormat(vcID string) (string, error) {
-	vcIDParts := strings.Split(vcID, "/")
+func (o *Operation) buildStructuredDoc(data *StoreVCRequest,
+	vc *verifiable.Credential) (*operation.StructuredDocument, error) {
+	var edvDocID string
 
-	uuidFromVCID := vcIDParts[len(vcIDParts)-1]
+	idFromMapping, err := o.idMapping.Get(vc.ID)
+	if err != nil {
+		edvDocID, err = generateEDVCompatibleID()
```
Baha-sk: we lost the original `err` here, are you sure you don't want to wrap/embed the 2 errors.. or at least log a warning with the original one?

since the original err is from a `Get()` function, I would assume you want to call `generateEDVCompatibleID` only if `Get()` returned a `notFound` error and other errors are valid error type that should return automatically .. 

this logic needs to be revisited.

test/bdd/fixtures/vc-rest/docker-compose.yml
```diff
@@ -17,6 +17,7 @@ services:
       - EDV_REST_HOST_URL=${EDV_HOST}:${EDV_PORT}
       - BLOC_DOMAIN=testnet.trustbloc.local:80
       - UNIVERSAL_RESOLVER_HOST_URL=http://uni-resolver-web:8080/1.0/identifiers
+      - LOCAL_STORE_PATH=bdd-test-local-store
```
fqutishat: ???

cmd/vc-rest/startcmd/start.go
fqutishat: you need to change this method

pkg/restapi/vc/operation/operations.go
```diff
@@ -322,41 +361,90 @@ func (o *Operation) storeVCHandler(rw http.ResponseWriter, req *http.Request) {
...
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		edvDocID = string(idFromMapping)
+	}
 
-	parsedUUID, err := uuid.Parse(uuidFromVCID)
+	doc := operation.StructuredDocument{}
+	doc.ID = edvDocID
```
Baha-sk: Shouldn't there be a check here `if edvDocID == ""` in case err was nil but edvDocID returned was empty above ? 

I guess this is another reason this logic needs to be revisited.

cmd/vc-rest/startcmd/start.go
DRK3: Oh, yes you're right. In fact, there's already a struct called vcRestParameters that's a perfect fit already. In hindsight it seems obvious...not sure why I missed that before. I've updated the code.

cmd/vc-rest/startcmd/start.go
DRK3: Oh, yes you're right. In fact, there's already a struct called vcRestParameters that's a perfect fit already. In hindsight it seems obvious...not sure why I missed that before. I've updated the code.

pkg/restapi/vc/operation/operations.go
DRK3: Gotcha. I've updated the code!

pkg/restapi/vc/operation/operations.go
DRK3: I see what you mean! I've updated the code.

pkg/restapi/vc/operation/operations.go
DRK3: Ah, when I first did this I had falsely assumed that the Aries storage didn't have a generic "not found" error, but now that I take a closer look it does indeed! I've updated the code to check for the specific condition now.

pkg/restapi/vc/operation/operations.go
DRK3: I  don't believe it's possible for edvDocID to be blank at this point, unless someone manually put a blank string in the database. But even if it was blank, the EDV would just return an error, which would cause the store VC handler to return an error back to the person who called that endpoint.

cmd/vc-rest/startcmd/start.go
DRK3: VCs have IDs and EDV encrypted documents have IDs. When someone calls a REST endpoint in edge-service, they will provide the VC ID, however this will not be the same as the ID of the associated encrypted document that contains the encrypted VC. So I need to map from VC ID -> EDV encrypted document ID, and I need to store that somewhere. So I store that in a local LevelDB database. This flag specifies where the database is created in the file system.

(It's like the [db-path flag in Aries-Framework-Go)](https://github.com/hyperledger/aries-framework-go/blob/6f509cae073a587e086972aaa93b9a878f88dfc2/cmd/aries-agent-rest/startcmd/start.go#L50)

test/bdd/fixtures/vc-rest/docker-compose.yml
DRK3: See my response to your other comment. With that in mind: within the docker container it doesn't really matter where the storage is for these tests, as long as it's writable.

pkg/restapi/vc/operation/operations.go
```diff
@@ -97,12 +112,33 @@ func New(provider storage.Provider, client Client, kms legacykms.KMS, vdri vdria
...
+	}
+
+	packer := authcrypt.New(kmsProv)
+
+	_, senderKey, err := kms.CreateKeySet()
+	if err != nil {
+		return nil, err
+	}
+
+	localStoreProvider := leveldb.NewProvider(localStorePath)
```
fqutishat: use provider.CreateStore instead of leveldb

fqutishat: similar to this err := provider.CreateStore(credentialStoreName)

fqutishat: no need for localStorePath just use localStoreName

test/bdd/fixtures/vc-rest/docker-compose.yml
fqutishat: no need

pkg/restapi/vc/operation/operations.go
```diff
@@ -20,11 +21,14 @@ import (
...
 	"github.com/google/uuid"
 	"github.com/gorilla/mux"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/packer/legacy/authcrypt"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/ed25519signature2018"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/verifiable"
 	vdriapi "github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdri"
 	"github.com/hyperledger/aries-framework-go/pkg/kms/legacykms"
+	ariesstorage "github.com/hyperledger/aries-framework-go/pkg/storage"
+	"github.com/hyperledger/aries-framework-go/pkg/storage/leveldb"
```
fqutishat: you don't need to use ariesstorage use github.com/trustbloc/edge-core/pkg/storage

cmd/vc-rest/startcmd/start.go
fqutishat: no need to use LOCAL_STORE_PATH follow credential store way

fqutishat:
@Baha-sk are you ok using ariesstorage????

cmd/vc-rest/startcmd/start_test.go
```diff
@@ -215,6 +195,41 @@ func TestCreateKMS(t *testing.T) {
...
+	require.NoError(t, err)
+
+	err = os.Unsetenv(edvURLEnvKey)
+	require.NoError(t, err)
+
+	err = os.Unsetenv(blocDomainEnvKey)
+	require.NoError(t, err)
+}
+
+func checkFlagPropertiesCorrect(t *testing.T, cmd *cobra.Command, flagName, flagShorthand, flagUsage string) {
```
DRK3: I just moved this method down, so all the Test methods could be together.

pkg/restapi/vc/operation/operations.go
```diff
@@ -130,17 +170,20 @@ type Operation struct {
...
 }
 
 func (o *Operation) createCredentialHandler(rw http.ResponseWriter, req *http.Request) {
 	data := CreateCredentialRequest{}
-	err := json.NewDecoder(req.Body).Decode(&data)
 
+	err := json.NewDecoder(req.Body).Decode(&data)
 	if err != nil {
-		o.writeErrorResponse(rw, http.StatusBadRequest, "Failed to write response for invalid request received")
+		o.writeErrorResponse(rw, http.StatusBadRequest, invalidRequestErrMsg)
```
DRK3: This error message was incorrect, so I fixed it.

pkg/restapi/vc/operation/operations.go
```diff
@@ -281,7 +324,7 @@ func (o *Operation) createProfileHandler(rw http.ResponseWriter, req *http.Reque
 	err := json.NewDecoder(req.Body).Decode(&data)
 
 	if err != nil {
-		o.writeErrorResponse(rw, http.StatusBadRequest, "Failed to write response for invalid request received")
+		o.writeErrorResponse(rw, http.StatusBadRequest, invalidRequestErrMsg)
```
DRK3: This error message was incorrect, so I fixed it.

pkg/restapi/vc/operation/operations.go
```diff
@@ -130,17 +170,20 @@ type Operation struct {
...
 }
 
 func (o *Operation) createCredentialHandler(rw http.ResponseWriter, req *http.Request) {
 	data := CreateCredentialRequest{}
-	err := json.NewDecoder(req.Body).Decode(&data)
 
+	err := json.NewDecoder(req.Body).Decode(&data)
 	if err != nil {
-		o.writeErrorResponse(rw, http.StatusBadRequest, "Failed to write response for invalid request received")
+		o.writeErrorResponse(rw, http.StatusBadRequest, fmt.Sprintf(invalidRequestErrMsg+": %s", err.Error()))
```
DRK3: This error message was incorrect, so I fixed it.

pkg/restapi/vc/operation/operations.go
```diff
@@ -278,16 +321,15 @@ func (o *Operation) updateCredentialStatusHandler(rw http.ResponseWriter, req *h
 
 func (o *Operation) createProfileHandler(rw http.ResponseWriter, req *http.Request) {
 	data := ProfileRequest{}
-	err := json.NewDecoder(req.Body).Decode(&data)
 
+	err := json.NewDecoder(req.Body).Decode(&data)
 	if err != nil {
-		o.writeErrorResponse(rw, http.StatusBadRequest, "Failed to write response for invalid request received")
+		o.writeErrorResponse(rw, http.StatusBadRequest, fmt.Sprintf(invalidRequestErrMsg+": %s", err.Error()))
```
DRK3: This error message was incorrect, so I fixed it.

pkg/restapi/vc/operation/operations.go
```diff
@@ -322,7 +364,7 @@ func (o *Operation) storeVCHandler(rw http.ResponseWriter, req *http.Request) {
 
 	err := json.NewDecoder(req.Body).Decode(&data)
 	if err != nil {
-		o.writeErrorResponse(rw, http.StatusBadRequest, "invalid request received")
+		o.writeErrorResponse(rw, http.StatusBadRequest, fmt.Sprintf(invalidRequestErrMsg+": %s", err.Error()))
```
DRK3: This error message was incorrect, so I fixed it.

pkg/restapi/vc/operation/operations.go
DRK3: Just as a heads up: this now uses the Value Not Found error from edge-core. Same idea. Works the same way.

pkg/restapi/vc/operation/operations_test.go
```diff
@@ -20,11 +20,16 @@ import (
...
 
+	"github.com/btcsuite/btcutil/base58"
+
 	"github.com/gorilla/mux"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/verifiable"
+	"github.com/hyperledger/aries-framework-go/pkg/kms/legacykms"
 	kmsmock "github.com/hyperledger/aries-framework-go/pkg/mock/kms/legacykms"
 	vdrimock "github.com/hyperledger/aries-framework-go/pkg/mock/vdri"
+	ariesstorage "github.com/hyperledger/aries-framework-go/pkg/storage"
```
DRK3: Aries storage is used here for manipulating KMS objects which exist in some tests. I believe the edge-core storage technically has the same signature, so they might appear to be compatible interfaces but the semantics are different due to the separate of Open/Create store in the edge-core version.

pkg/restapi/vc/operation/operations_test.go
```diff
@@ -1573,3 +1709,34 @@ func (m *mockVCStatusManager) UpdateVCStatus(v *verifiable.Credential, profile *
 func (m *mockVCStatusManager) GetCSL(id string) (*cslstatus.CSL, error) {
 	return m.getCSLValue, m.getCSLErr
 }
+
+// FailOnSecondOpenMockProvider is a mock provider that returns an err the second time the OpenStore method is called.
+type FailOnSecondOpenMockProvider struct {
```
DRK3: Interestingly enough, I actually have a case where I needed it to fail the second time...  (I remember I asked about this hypothetical situation awhile back @sandrask, but I didn't think I'd actually see it so soon)

PR #126, opened 2020-03-19
feat: KMS Generate Keypair API

### LINKED ISSUE #123: Generate Keypair API
The issue credential api #116 expects the private key associated with the DID to be available in the KMS of the issuer service. 

One of the ways to achieve that is to have the issuer-service generate the keypair used by the DID. 
1. generate the keypair inside the service.
2. Create DID Document using the previously generate keys (outside of the service).
3. Call issue credential API with DID for signing (#116).

rolsonquadras:
- New API to generate keypair from KMS
- Moved getPublicKey from DID Document logic to a common function

closes #123 

### DIFF
--- a/pkg/restapi/vc/operation/operations.go
+++ b/pkg/restapi/vc/operation/operations.go
@@ -55,6 +55,8 @@ const (
 	vcStatusEndpoint               = vcStatus + "/{id}"
 	credentialsBasePath            = "/credentials"
 	issueCredentialPath            = credentialsBasePath + "/issueCredential"
+	kmsBasePath                    = "/kms"
+	generateKeypairPath            = kmsBasePath + "/generatekeypair"
 
 	successMsg = "success"
 	cslSize    = 50
@@ -683,15 +685,9 @@ func (o *Operation) createProfile(pr *ProfileRequest) (*vcprofile.DataProfile, e
 		}
 	}
 
-	var publicKeyID string
-
-	switch {
-	case len(didDoc.PublicKey) > 0:
-		publicKeyID = didDoc.PublicKey[0].ID
-	case len(didDoc.Authentication) > 0:
-		publicKeyID = didDoc.Authentication[0].PublicKey.ID
-	default:
-		return nil, fmt.Errorf("can't find public key in DID")
+	publicKeyID, err := getPublicKeyID(didDoc)
+	if err != nil {
+		return nil, err
 	}
 
 	created := time.Now().UTC()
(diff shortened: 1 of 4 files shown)


pkg/restapi/vc/operation/operations_test.go
```diff
@@ -1551,7 +1551,7 @@ func TestIssueCredential(t *testing.T) {
 
 		op.didBlocClient = &didbloc.Client{CreateDIDValue: didDoc}
 
-		issueCredentialHandler := getHandler(t, op, issueCredentialPath, mode)
+		issueCredentialHandler := getHandler(t, op, issueCredentialPath, issuerMode)
```
sandrask: It makes more sense to change name of issuerCredentialHandler to credentialHandler. Edge service operates in two modes: "issuer" and "verifier" so issuerMode doesn't make sense.

rolsonquadras: it's "issueCredentialHandler" (name based on issue credential API) not "issuerCredentialHandler".

pkg/restapi/vc/operation/operations.go
```diff
@@ -55,6 +55,8 @@ const (
 	vcStatusEndpoint               = vcStatus + "/{id}"
 	credentialsBasePath            = "/credentials"
 	issueCredentialPath            = credentialsBasePath + "/issueCredential"
+	kmsBasePath                    = "/kms"
```
llorllale: nit: not sure this name is appropriate given that this system is not a KMS

pkg/restapi/vc/operation/operations.go
```diff
@@ -861,6 +851,21 @@ func (o *Operation) issueCredentialHandler(rw http.ResponseWriter, req *http.Req
...
+	if err != nil {
+		o.writeErrorResponse(rw, http.StatusInternalServerError,
+			fmt.Sprintf("failed to create key pair: %s", err.Error()))
+
+		return
+	}
+
+	rw.WriteHeader(http.StatusOK)
+	o.writeResponse(rw, &GenerateKeyPairResponse{
+		PublicKey: signKey,
```
llorllale: Can this be a handle instead?

pkg/restapi/vc/operation/operations.go
rolsonquadras: well, its not domain name - its subpath (operation on kms). Ideally, full path would look like "https://issuer.example.com/kms/generatekeypair".

pkg/restapi/vc/operation/operations.go
rolsonquadras: Not sure I understand your comment. Can you please elaborate ?

llorllale: I was referring to a key handle (some opaque reference, eg. uuid). Instead of returning the raw public key contents, can it instead return a reference?

rolsonquadras: The consumer would need the actual public key to create the DID Document.

pkg/restapi/vc/operation/operations.go
llorllale: a path named "kms" would denote a KMS resource.

But I'm not going to be picky about this one - we can easily refactor later.

rolsonquadras: Agree on the refactoring part.

PR #139, opened 2020-03-25
feat: support for did element

### LINKED ISSUE #138: Interop Test: DID Element

sudeshrshetty:
- addes support for did element and tests
- closes #138

Signed-off-by: sudesh.shetty <sudesh.shetty@securekey.com>


fqutishat:
@sudeshrshetty  after this PR is merged we need to add new profile in sandbox https://github.com/trustbloc/edge-sandbox/blob/master/test/bdd/fixtures/scripts/vcs_configure.sh#L42

https://github.com/trustbloc/edge-sandbox/blob/master/cmd/issuer-rest/static/index.html#L39

sudeshrshetty:
> @sudeshrshetty after this PR is merged we need to add new profile in sandbox https://github.com/trustbloc/edge-sandbox/blob/master/test/bdd/fixtures/scripts/vcs_configure.sh#L42
> 
> https://github.com/trustbloc/edge-sandbox/blob/master/cmd/issuer-rest/static/index.html#L39

Sure

pkg/restapi/vc/operation/operations.go
```diff
@@ -990,16 +999,29 @@ func (o *Operation) parseAndVerifyVP(vpBytes []byte) (*verifiable.Presentation,
...
+		var publicKeyID string
+
+		for _, k := range didDoc.PublicKey {
+			if strings.HasPrefix(k.Type, Ed25519VerificationKey) {
+				publicKeyID = k.ID
+				break
+			}
+		}
+
+		if !isDID(publicKeyID) {
```
sandrask: So their DID keys are not URIs (not as per spec). Can you add comment (or create issue) to remove this condition when they become compliant?

sudeshrshetty: DID element has existing issue to support public keys in with DID based ids. Created #140 
 and added a TODO to remove it in future.

sudeshrshetty:
> I'm not a golang expert, but the JS parts look correct. Great work!

Thanks @OR13

pkg/restapi/vc/operation/operations.go
OR13: They should be compliant now: https://github.com/decentralized-identity/element/pull/179

https://uniresolver.io/1.0/identifiers/did:elem:EiBJJPdo-ONF0jxqt8mZYEj9Z7FbdC87m2xvN0_HAbcoEg

PR #155, opened 2020-03-27
refactor: Issuer API BDD tests

rolsonquadras:
- Added scenario for issueCred to be inline with composeAndIssueCred


test/bdd/features/issuer_e2e_api.feature
```diff
@@ -10,11 +10,16 @@ Feature: Issuer VC REST API
...
-    Then  Verify the proof value generated using the Issuer Service Issue Credential API with the DID stored in "did" variable
+    Given "University" has a DID
+    Then  "University" application service verifies the credential created by Issuer Service issueCredential API with it's DID
 
   @composeAndIssueCred_api
   Scenario: Compose and Issue Credential API
-    Given "University" has a did
-    Then  "University" application service verifies the credential created by Issuer Service Issue Credential API
+    Given "University" has a DID
+    Then  "University" application service verifies the credential created by Issuer Service composeAndIssueCredential API with it's DID
```
sandrask: we should not be using composeAndIssueCredential in BDD test - it should be human readable. 

Given "University" has a DID
Then  "University" application service verifies that the credential created by Issuer Service is created with it's DID

test/bdd/pkg/issuer/issuer_steps.go
```diff
@@ -254,6 +259,12 @@ func (e *Steps) signCredential(didDocVar string) ([]byte, error) {
 	return responseBytes, nil
 }
 
+func (e *Steps) issueCred(user string) error {
```
sandrask: It is funny that issueCred calls verifyCredential only

test/bdd/features/issuer_e2e_api.feature
rolsonquadras: I'll update in next PR.

test/bdd/pkg/issuer/issuer_steps.go
rolsonquadras: I'll update in next PR.

PR #218, opened 2020-04-08
chore: Update HL Aries and trustbloc-did dependency version

rolsonquadras:


fqutishat:
@rolsonquadras  you need to update https://github.com/trustbloc/edge-service/blob/master/test/bdd/fixtures/did-method-rest/.env#L8

PR #252, opened 2020-04-14
chore: Make credential status optional

### LINKED ISSUE #248: Make credential status optional
- disable credential status for https://issuer.sandbox.trustbloc.dev/vc-issuer-1/
 (do not output)

fqutishat:
closes #248

### DIFF
--- a/pkg/doc/vc/profile/profile.go
+++ b/pkg/doc/vc/profile/profile.go
@@ -40,6 +40,7 @@ type DataProfile struct {
 	Creator                 string                             `json:"creator"`
 	Created                 *time.Time                         `json:"created"`
 	DIDPrivateKey           string                             `json:"didPrivateKey"`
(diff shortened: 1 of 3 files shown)


pkg/doc/vc/profile/profile.go
```diff
@@ -40,6 +40,7 @@ type DataProfile struct {
 	Creator                 string                             `json:"creator"`
 	Created                 *time.Time                         `json:"created"`
 	DIDPrivateKey           string                             `json:"didPrivateKey"`
+	DisableVCStatus         bool                               `json:"disableVCStatus"`
```
troyronda, line 43: unrelated to this PR, but we should align on underscore JSON keys (like Sidetree).
https://github.com/trustbloc/edge-service/issues/254

troyronda, line 43: note: I think it would be better to think about this as the default behaviour for this profile rather than as "Disable"

PR #276, opened 2020-04-16
feat: Use EDV encrypted index for ID mapping

### LINKED ISSUE #112: Use EDV Encrypted Indices instead of a mapping for IDs
By having the ID mapping in the same provider, the VC ID (which is also present in the encrypted document) is being made available to the provider, and so this represents a leak of the encrypted data. Use EDV Encrypted Indices instead, and get rid of the ID map.

DRK3:

closes #112


pkg/restapi/vc/operation/operations_test.go
```diff
@@ -1013,49 +1146,41 @@ func TestRetrieveVCHandler(t *testing.T) {
...
 		q.Add("profile", getTestProfile().Name)
-		req.URL.RawQuery = q.Encode()
+		r.URL.RawQuery = q.Encode()
+		rr := httptest.NewRecorder()
 
-		rw := mockResponseWriter{}
-		retrieveVCHandler.Handle().ServeHTTP(rw, req)
-		require.Contains(t, logContents.String(),
-			"Failed to write response for document retrieval success: response writer failed")
-	})
```
DRK3, line 1047: This was a duplicate test, so I removed it

pkg/restapi/vc/operation/operations.go
```diff
@@ -1324,6 +1325,117 @@ func (o *Operation) parseAndVerifyVP(vpBytes []byte) error {
...
+
+func (o *Operation) retrieveCredential(rw http.ResponseWriter, profileName string, docIDs []string) {
+	var retrievedVC []byte
+
+	switch len(docIDs) {
+	case 0:
+		o.writeErrorResponse(rw, http.StatusBadRequest,
+			fmt.Sprintf(`no VC under profile "%s" was found with the given id`, profileName))
+	case 1:
+		splitBySlashes := strings.Split(docIDs[0], `/`)
```
fqutishat, line 1350: can you explain logic here

DRK3: Which logic?

DRK3: I see the confusion. The EDV query vault method returns the full URLs of the document (per the spec), and I forgot to update some of the variable names here,

I've updated the PR now to fix the variable names and I also moved the split by slash stuff into a separate function. It should be more clear now.

pkg/restapi/vc/operation/operations.go
```diff
@@ -1324,6 +1325,115 @@ func (o *Operation) parseAndVerifyVP(vpBytes []byte) error {
...
+		// Multiple VCs were found with the same id. This is technically possible under the right circumstances
+		// when storing the same VC multiples times in a store provider that follows an "eventually consistent"
+		// consistency model. If they are all the same, then just return the first one arbitrarily.
+		// ToDo: If the multiple VCs with the same ID all are identical then delete the extras and leave only one.
+		// https://github.com/trustbloc/edge-service/issues/262
+		var err error
+
+		retrievedVC, err = o.verifyMultipleMatchingVCsAreIdentical(profileName, docURLs)
+		if err != nil {
+			o.writeErrorResponse(rw, http.StatusInternalServerError, err.Error())
```
bstasyszyn, line 1370: I don't think internal server error is appropriate here. We could maybe use 409 - CONFLICT: 

> The request could not be completed due to a conflict with the current state of the target resource. This code is used in situations where the user might be able to resolve the conflict and resubmit the request.

DRK3: Note that the `verifyMultipleMatchingVCsAreIdentical` method can also return an error if the `retrieveVC` method fails, and I think failures from `retrieveVC` are still internal server errors. So I've made `verifyMultipleMatchingVCsAreIdentical` return a status code now depending on which failure - if there are multiple VCs with differing contents, I return the 409 Conflict error now. Let me know if you disagree with my change.

Issue #285, opened 2020-04-17
Configure plugfest test with at least 2 different DID methods as an issuer


troyronda:
3.2 Issuer must support issuance of credentials with at least 2 different DID methods as an issuer
          ✕ meets criteria after all issuers have been tested (2ms)

----

        10. Verifier must be able to verify credentials with at least 2 different DID methods as the did issuer
          ✕ should pass (143ms)

troyronda:
create issuer profile for each interop did:method. Leave `vc-issuer-interop` as did:trustbloc

e.g., vc-issuer-interop-v1, vc-issuer-interop-sov, vc-issuer-interop-key, ...

troyronda:
- configure https://github.com/w3c-ccg/vc-examples/tree/master/plugfest-2020/vendors/trustbloc
- ensure above tests pass.

rolsonquadras:
Need to update the test suite configuration after next deployment.

rolsonquadras:
The second issuer has been pushed to W3C test suite. Here is the latest W3C interop report - https://w3c-ccg.github.io/vc-examples/plugfest-2020.html

PR #297, opened 2020-04-17
refactor: Search private key from profile if verificationMethod matches the profile DID

rolsonquadras:
part of #285 


pkg/doc/vc/crypto/crypto.go
```diff
@@ -181,7 +181,19 @@ func (c *Crypto) SignCredential(dataProfile *vcprofile.DataProfile, vc *verifiab
...
 	switch {
 	case opts.VerificationMethod != "":
+		did, err := getDIDFromKeyID(opts.VerificationMethod)
+		if err != nil {
+			return nil, "", err
+		}
+
+		// if the verification method DID is added to profile externally, then fetch the private
+		// key from profile
+		if did == dataProfile.DID && dataProfile.DIDPrivateKey != "" {
```
sudeshrshetty, line 191: this may not work when someone passes verification method through options and doesn't want to use DID/privatekey from profile

PR #359, opened 2020-04-30
refactor: Crypto setup code

DRK3:

### DIFF
--- a/pkg/restapi/vc/operation/operations.go
+++ b/pkg/restapi/vc/operation/operations.go
@@ -166,13 +166,14 @@ func New(config *Config) (*Operation, error) {
 		return nil, fmt.Errorf("failed to instantiate new csl status: %w", err)
 	}
 
-	jweEncrypter, jweDecrypter, err := cryptosetup.PrepareJWECrypto(config.KeyManager, config.StoreProvider)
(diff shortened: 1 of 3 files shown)


pkg/restapi/vc/operation/operations.go
```diff
@@ -166,13 +166,14 @@ func New(config *Config) (*Operation, error) {
 		return nil, fmt.Errorf("failed to instantiate new csl status: %w", err)
 	}
 
-	jweEncrypter, jweDecrypter, err := cryptosetup.PrepareJWECrypto(config.KeyManager, config.StoreProvider)
+	jweEncrypter, jweDecrypter, err := cryptosetup.PrepareJWECrypto(config.KeyManager, config.StoreProvider,
+		jose.A256GCM, kms.ECDHES256AES256GCMType)
```
DRK3, line 170: I also noticed that I was using `kms.ECDHES256AES256GCM` before instead of `kms.ECDHES256AES256GCMType`, so I fixed that (although I suppose they work the same anyway, I think you intended for us to use `kms.ECDHES256AES256GCMType` @Baha-sk?

Moopli: FooType vs Foo is simply a custom type for code clarity, so yes :+1:

PR #404, opened 2020-05-05
fix: Add retrieve VC retries

DRK3:

### DIFF
--- a/cmd/did-rest/go.mod
+++ b/cmd/did-rest/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/sirupsen/logrus v1.4.2
 	github.com/spf13/cobra v0.0.6
 	github.com/stretchr/testify v1.5.1
-	github.com/trustbloc/edge-core v0.1.3
+	github.com/trustbloc/edge-core v0.1.4-0.20200514152520-6eccfe31d648
 	github.com/trustbloc/edge-service v0.0.0
 )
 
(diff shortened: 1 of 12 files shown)


pkg/restapi/vc/operation/operations.go
```diff
@@ -689,11 +689,27 @@ func (o *Operation) retrieveCredentialHandler(rw http.ResponseWriter, req *http.
...
 
-	docURLs, err := o.queryVault(profile, id)
-	if err != nil {
-		o.writeErrorResponse(rw, http.StatusInternalServerError, err.Error())
+	var docURLs []string
 
-		return
+	for retries := 0; retries < 10; retries++ {
+		if retries != 0 {
+			time.Sleep(time.Second * 5)
```
sandrask, line 696: 5 seconds is too long in between first and second retry

troyronda: @bstasyszyn

troyronda: also 50 seconds maximum is too much :).

bstasyszyn: For example, see: https://github.com/trustbloc/fabric-mod/blob/ba9a41234a5002930b1232c4640b86fbd1bf9bb7/common/util/retry/retry.go#L47

DRK3: Reworked it - take a look and let me know.

pkg/restapi/vc/operation/operations.go
```diff
@@ -693,11 +693,32 @@ func (o *Operation) retrieveCredentialHandler(rw http.ResponseWriter, req *http.
 		return
 	}
 
-	docURLs, err := o.queryVault(profile, id)
-	if err != nil {
-		o.writeErrorResponse(rw, http.StatusInternalServerError, err.Error())
+	var docURLs []string
 
-		return
+	backoff := 250 * time.Millisecond
```
bstasyszyn, line 698: These values (initial backoff, maxRetries, and backoff factor) should be configurable.

fqutishat: you should add retry logic outside operation as common library in edge-core

DRK3: Done - and it's being used here now.

PR #405, opened 2020-05-05
feat: Support for DID Factom

rolsonquadras:

### DIFF
--- a/cmd/vc-rest/startcmd/start.go
+++ b/cmd/vc-rest/startcmd/start.go
@@ -140,6 +140,7 @@ const (
 	didMethodSov     = "sov"
 	didMethodWeb     = "web"
 	didMethodKey     = "key"
+	didMethodFactom  = "factom"
 
 	masterKeyURI = "local-lock://custom/master/key/"
 )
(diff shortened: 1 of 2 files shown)


cmd/vc-rest/startcmd/start.go
```diff
@@ -501,7 +502,7 @@ func supportedMode(mode string) bool {
 // acceptsDID returns if given did method is accepted by VC REST api
 func acceptsDID(method string) bool {
 	return method == didMethodVeres || method == didMethodElement || method == didMethodSov ||
-		method == didMethodWeb || method == didMethodKey
+		method == didMethodWeb || method == didMethodKey || method == didMethodFactom
```
troyronda, line 505: I'm somewhat curious why we don't just reply true :)

troyronda: do we have multiple DID acceptors in this service?
if we need to do it this way, we need a followup to make this configurable.

rolsonquadras: May be original idea was to have some whitelist. Probably we can move this to a config.

rolsonquadras: https://github.com/trustbloc/edge-service/issues/406

PR #413, opened 2020-05-08
chore: Import private keys into kms

fqutishat:
cloes #400


pkg/restapi/vc/operation/operations.go
```diff
@@ -731,40 +734,70 @@ func (o *Operation) createDIDUniRegistrar(keyType, signatureType, purpose string
...
+	return identifier, keys[0].ID, nil
+}
+
+func (o *Operation) importKey(keyID string, keyType kms.KeyType, privateKeyBytes []byte) error {
+	split := strings.Split(keyID, "#")
+
+	var privKey interface{}
+
+	switch keyType {
+	case kms.ED25519Type:
```
sandrask, line 785: Is import only supported for Ed25519 for now or forever?

pkg/restapi/vc/operation/operations.go
```diff
@@ -731,40 +734,70 @@ func (o *Operation) createDIDUniRegistrar(keyType, signatureType, purpose string
...
+		privKey = ed25519.PrivateKey(privateKeyBytes)
+	default:
+		return fmt.Errorf("import key type not supported %s", keyType)
+	}
+
+	_, _, err := o.kms.ImportPrivateKey(privKey,
```
sandrask, line 791: Are we going to have key id collisions? For example in Sidetree protocol key id can be an arbitrary name.

pkg/restapi/vc/operation/operations.go
```diff
@@ -849,50 +881,49 @@ func (o *Operation) createHolderProfile(pr *HolderProfileRequest) (*vcprofile.Ho
...
-			return "", "", "", fmt.Errorf("failed to resolve did: %v", err)
+			return "", "", fmt.Errorf("failed to resolve did: %v", err)
 		}
 
 		didID = didDoc.ID
 
-		publicKeyID, err = getPublicKeyID(didDoc, "", crypto.Ed25519Signature2018)
-		if err != nil {
-			return "", "", "", err
+		if err := o.importKey(keyID, kms.ED25519Type, base58.Decode(privateKey)); err != nil {
```
sandrask, line 919: Are we assuming that if did is there that key has to be Ed25519 type? This can be handled in separate PR - it is an existing issue.

pkg/restapi/vc/operation/operations.go
fqutishat: we don't control the id

fqutishat: but collisions is  possible

fqutishat: we need to revisit

pkg/restapi/vc/operation/operations.go
fqutishat: for now yes because all vendors now just support ed25519

pkg/restapi/vc/operation/operations.go
sandrask: hopefully KMS will reject it, otherwise we may overwrite someone's key

pkg/restapi/vc/operation/operations.go
fqutishat: need to change when vendor start support multiple keys

pkg/restapi/vc/operation/operations.go
fqutishat: for now

pkg/restapi/vc/operation/operations.go
sandrask: Since KMS will reject import if key already exists we should be safe here

PR #446, opened 2020-05-20
feat: add native support for did:key in did-rest

aholovko:
Adds support for did:key in did-rest using key VDRI from aries-framework-go.

Use the following configuration to enable VDRI implementation:

```json
{
  "pattern": "^(did:key:.+)$"
}
```
Note: Put the rule above **before** "catch-all" rule for uniresolver.

Signed-off-by: Andriy Holovko <andriy.holovko@gmail.com>


troyronda:
@cla-bot check

pkg/restapi/did/controller_test.go
```diff
@@ -38,10 +39,10 @@ type mockRuleProvider struct {
 	Err error
 }
 
-func (p *mockRuleProvider) Transform(uri string) (string, error) {
+func (p *mockRuleProvider) Resolve(uri string) (rules.ResolveResult, error) {
```
sandrask, line 42: I think it is a bit confusing for rule provider to have Resolve method (it will be confused with dids) - maybe Match makes more sense if you don't like original Transform.

pkg/restapi/did/operation/operations.go
```diff
@@ -72,41 +79,66 @@ type httpClient interface {
...
-// Proxy will proxy requests based on provided configuration
-func (o *Operation) proxy(rw http.ResponseWriter, req *http.Request) {
+func (o *Operation) resolve(rw http.ResponseWriter, req *http.Request) {
 	did := mux.Vars(req)["did"]
 
-	log.Debugf("proxy received request for DID: %s", did)
+	log.Debugf("resolve received request for DID: %s", did)
 
-	destinationURL, err := o.ruleProvider.Transform(did)
+	result, err := o.ruleProvider.Resolve(did)
```
sandrask, line 91: Already here having Resolve is confusing - I would be better to change it to Match (or keep Transform)

pkg/restapi/did/operation/operations.go
```diff
@@ -72,41 +79,66 @@ type httpClient interface {
...
+		return
+	}
+
+	bytes, err := doc.JSONBytes()
+	if err != nil {
+		writeBadRequestResponse(rw, fmt.Sprintf("failed to convert DIDDoc to json bytes: %s", err.Error()))
+		return
+	}
+
+	rw.Header().Add(contentTypeHeader, didLDJson)
```
sandrask, line 118: Code from here to the end of the function is same as in resolveWithProxy - you should extract it into separate function for both calls.

pkg/restapi/did/operation/operations.go
```diff
@@ -145,21 +177,18 @@ func addRequestHeaders(req, newReq *http.Request) {
 	}
 }
 
-func (o *Operation) writeErrorResponse(rw http.ResponseWriter, status int, msg string) {
-	log.Warnf("proxy returning status code: %d, error: %s", status, msg)
+func writeBadRequestResponse(rw http.ResponseWriter, msg string) {
```
sandrask, line 180: No reason to change this - you may want to use it for status other than bad request

pkg/restapi/did/operation/operations.go
```diff
@@ -72,41 +79,66 @@ type httpClient interface {
...
+func (o *Operation) resolveWithVDRI(rw http.ResponseWriter, did string) {
+	doc, err := o.vdri.Resolve(did)
+	if err != nil {
+		writeBadRequestResponse(rw, fmt.Sprintf("failed to resolve DID: %s", err.Error()))
+		return
+	}
+
+	bytes, err := doc.JSONBytes()
+	if err != nil {
+		writeBadRequestResponse(rw, fmt.Sprintf("failed to convert DIDDoc to json bytes: %s", err.Error()))
```
sandrask, line 114: Is this bad request? This error is more like internal server error.

pkg/restapi/did/operation/operations.go
```diff
@@ -145,12 +165,22 @@ func addRequestHeaders(req, newReq *http.Request) {
 	}
 }
 
-func (o *Operation) writeErrorResponse(rw http.ResponseWriter, status int, msg string) {
+func writeOKResponse(rw http.ResponseWriter, status int, contentType string, body []byte) {
```
sandrask, line 168: Rename to writeResponse since it takes status code (may not be OK)

cmd/did-rest/startcmd/test/config.json
```diff
@@ -4,6 +4,10 @@
       "pattern": "^(did:trustbloc:testnet.trustbloc.local:.+)$",
       "url": "http://trustbloc.did.method.example.com:8060/resolveDID?did=$1"
     },
+    {
+      "pattern": "^(did:key:.+)$",
+      "vdri": true
```
troyronda, line 9: I wonder if this is the most descriptive configuration property?

thoughts? @fqutishat @sandrask @bstasyszyn @aholovko

cmd/did-rest/startcmd/start.go
```diff
@@ -188,6 +198,24 @@ func startDidService(parameters *didRestParameters, srv server) error {
...
+	var opts []vdripkg.Option
+
+	keyVDRI, err := key.New()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create key vdri: %w", err)
+	}
+
+	opts = append(opts, vdripkg.WithVDRI(keyVDRI))
+
+	vdriProvider, err := context.New(context.WithLegacyKMS(nil))
```
sandrask, line 211: Do you need legacy KMS here?

cmd/did-rest/startcmd/test/config.json
sandrask: It is descriptive from developer point since we use VDRI :) How about embedded or local? Not sure those names are any better.

pkg/restapi/did/operation/operations.go
aholovko: done

cmd/did-rest/startcmd/start.go
aholovko: removed :)

cmd/did-rest/startcmd/test/config.json
aholovko: I suppose that the "full" configuration will be something like this:
```json
{
  "pattern": "^(did:btcr:.+)$",
  "image": "universalresolver/driver-did-btcr",
  "imageProperties": "true",
  "tag": "latest",
  "testIdentifiers": ["did:btcr:xz35-jznz-q9yu-ply"]
}
```

Perhaps `"local":true` looks not so bad among other properties. It denotes that it's not the docker image but the local implementation.

fqutishat: we should not expose  "vdri": true

sandrask: @fqutishat we all agree on that - we are just trying to come up with meaningful name - I am ok with local @troyronda?

rolsonquadras:
@aholovko Here is the config for BDD tests - https://github.com/trustbloc/edge-service/blob/master/test/bdd/fixtures/did-rest/config.json#L1. Not sure if BDD tests are planned for another PR (just in case). 

Note - we are already using did:key for issuance/verifications.

aholovko:
> @aholovko Here is the config for BDD tests - https://github.com/trustbloc/edge-service/blob/master/test/bdd/fixtures/did-rest/config.json#L1. Not sure if BDD tests are planned for another PR (just in case).
> 
> Note - we are already using did:key for issuance/verifications.

Thanks @rolsonquadras for noticing that. Currently there is a problem with switching to vdri implementation of did key in BDD tests - the problem lies in `populateVerificationRelationships()` method of `did.Doc` ("keyAgreements" section). I don't know the all details yet but I am going to create ticket for that and let Dima knows about the issue.

cmd/did-rest/startcmd/test/config.json
troyronda: sure - local, embedded, system would be possibilities.

fqutishat: no field is needed just ```"pattern": "^(did:key:.+)$"```

troyronda: I suppose that's true. If url and image are not present, I guess it's implied that it is a built-in handler.

cmd/did-rest/startcmd/start.go
```diff
@@ -188,6 +198,24 @@ func startDidService(parameters *didRestParameters, srv server) error {
...
 
+func createVDRI() (vdriapi.Registry, error) {
+	var opts []vdripkg.Option
+
+	keyVDRI, err := key.New()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create key vdri: %w", err)
+	}
+
+	opts = append(opts, vdripkg.WithVDRI(keyVDRI))
```
llorllale, line 209: @aholovko any reason why it isn't already in the aries framework's defaults?

cmd/did-rest/startcmd/test/config.json
llorllale: I'm looking at this code for the first time.

To me it looks like `TransformResult` should be a resolver.

llorllale: If we remove all keys and just leave `pattern` in this config object, then are we left with "rules" or just a whitelist?

cmd/did-rest/startcmd/start.go
aholovko: @llorllale I think that we should do it. If everyone agrees I will create a ticket for that.

llorllale: @aholovko I agree

aholovko: @llorllale I created ticket [#1847](https://github.com/hyperledger/aries-framework-go/issues/1847) for adding key vdri to framework defaults, as it looks to be a useful functionality. But here, I will probably leave the existing logic as we are not using the "full" aries framework, only the vdri component.

Thank you @fqutishat again for clarifying that part to me.

pkg/doc/vc/status/csl/csl.go
```diff
@@ -190,7 +190,7 @@ func (c *CredentialStatusManager) createStatusCredential(v *verifiable.Credentia
 		return nil, fmt.Errorf("create credential marshalling failed: %s", err.Error())
 	}
 
-	validatedStatusCred, _, err := verifiable.NewCredential(cred)
+	validatedStatusCred, err := verifiable.ParseCredential(cred)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create new credential: %s", err.Error())
```
sudeshrshetty, line 195: error message still says new credential

pkg/doc/vc/status/csl/csl.go
```diff
@@ -190,7 +190,7 @@ func (c *CredentialStatusManager) createStatusCredential(v *verifiable.Credentia
 		return nil, fmt.Errorf("create credential marshalling failed: %s", err.Error())
 	}
 
-	validatedStatusCred, _, err := verifiable.NewCredential(cred)
+	validatedStatusCred, err := verifiable.ParseCredential(cred)
```
sudeshrshetty, line 193: What is the reason for moving from 'NewCredential' to 'ParseCredential'?

pkg/restapi/issuer/operation/operations.go
```diff
@@ -870,7 +870,7 @@ func (o *Operation) createKey(keyType kms.KeyType) (string, []byte, error) {
 }
 
 func (o *Operation) parseAndVerifyVC(vcBytes []byte) (*verifiable.Credential, error) {
-	vc, _, err := verifiable.NewCredential(
+	vc, err := verifiable.ParseCredential(
```
sudeshrshetty, line 873: Why 'ParseCredential'?

pkg/doc/vc/status/csl/csl.go
aholovko: no sure, these are sync up changes with Aries framework
> What is the reason for moving from 'NewCredential' to 'ParseCredential'?
@kdimak Can you plz check this

troyronda: this was refactored in aries and is correct.

pkg/doc/vc/status/csl/csl.go
aholovko: updated

PR #458, opened 2020-06-29
feat: Set log level via command line flag

DRK3:

### DIFF
--- a/cmd/did-rest/startcmd/start_test.go
+++ b/cmd/did-rest/startcmd/start_test.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/require"
+	"github.com/trustbloc/edge-core/pkg/log"
 )
 
 type mockServer struct{}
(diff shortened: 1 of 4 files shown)


cmd/did-rest/startcmd/start_test.go
```diff
@@ -13,6 +13,8 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/require"
+
```
rolsonquadras, line 16: extra line

cmd/vc-rest/startcmd/start_test.go
```diff
@@ -13,6 +13,8 @@ import (
 	"os"
 	"testing"
 
+	"github.com/trustbloc/edge-core/pkg/log"
+
```
rolsonquadras, line 17: extra line

cmd/did-rest/startcmd/start_test.go
DRK3: Fixed - good catch.

cmd/vc-rest/startcmd/start_test.go
DRK3: Fixed - good catch.

PR #478, opened 2020-08-24
chore: update aries-framework-go - remove LegacyKMS

Baha-sk:


Baha-sk:
> Your BDD tests are failing, start with _ issue

I was wondering what can be done since kms KIDs are now JWK thumbprint for asymmetric keys, I can't tweak the key as it will break the thumbprint value. Shall I encode the kid as base58 instead of base64URL Raw encoding (no padding)?

sandrask:
For DCAS we did conversion internally from Base64 to Base58 before storing/retrieving

Write(doc) returns Base64 ID
- calculate Base64 ID from document 
- convert Base64 ID to  Base58 ID
- store document with Base58 ID 
- return Base64 ID

Read(Base64 id): 
- convert Base64 ID to Base58 ID
- retrieve document based on Base58 ID

Baha-sk:
> @Baha-sk For DCAS we did conversion internally from Base64 to Base58 before storing/retrieving
> 
> Write(doc) returns Base64 ID
> 
> * calculate Base64 ID from document
> * convert Base64 ID to  Base58 ID
> * store document with Base58 ID
> * return Base64 ID
> 
> Read(Base64 id) returns document
> 
> * convert Base64 ID to Base58 ID
> * retrieve document based on Base58 ID

@sandrask indeed,  I also wrote a small app to generate 1000 keys and noticed base64 encoding do have `_` while converted base58 encoding don't.

I guess since the KMS uses a storage for the keys, I need to use base58 encoded `KID`s since it will fail when using a storage provider that reserves IDs starting with `_`.

I will update the KMS to use base58 encoded `KID`s as I can't alter the jwk thumbprint []byte value. I think converting the thumbprint between base64 and base58 will confuse the user. We'll just stick with base58 encoded IDs in the KMS for consistency's sake.

sandrask:
@Baha-sk CAS ID had to be Base64 encoded so we couldn't use Base58 as ID that is returned to the client.

Baha-sk:
@sandrask with the latest Aries-framework-go commit (as of today) a new store wrapper is added that wraps base64 encoded keys into base58 keys to fix this issue. BDD tests should not fail on this now.

PR #516, opened 2020-10-26
feat: delete verifier profile api

### LINKED ISSUE #503: Verifier Profile : Delete API
- [ ] New endpoint for delete profile
- [ ] BDD tests to show create, delete and create api call (profile recreation).

grace0925:
closes #503 
Signed-off-by: Grace Liu <grace.liu@securekey.com>

### DIFF
--- a/pkg/restapi/verifier/operation/operations.go
+++ b/pkg/restapi/verifier/operation/operations.go
@@ -37,10 +37,12 @@ const (
 	verifierBasePath                  = "/verifier"
 	profileEndpoint                   = verifierBasePath + "/profile"
 	getProfileEndpoint                = profileEndpoint + "/" + "{" + profileIDPathParam + "}"
+	deleteProfileEndpoint             = profileEndpoint + "/" + "{" + profileIDPathParam + "}"
 	credentialsVerificationEndpoint   = "/" + "{" + profileIDPathParam + "}" + verifierBasePath + "/credentials"
 	presentationsVerificationEndpoint = "/" + "{" + profileIDPathParam + "}" + verifierBasePath + "/presentations"
 
-	invalidRequestErrMsg = "Invalid request"
+	invalidRequestErrMsg          = "Invalid request"
+	verifierProfileNotFoundErrMsg = "Verifier profile with id %s does not exist: %s"
 
 	successMsg = "success"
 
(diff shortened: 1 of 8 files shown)


fqutishat (approved):

pkg/restapi/verifier/operation/operations.go
```diff
@@ -182,6 +186,35 @@ func (o *Operation) getProfileHandler(rw http.ResponseWriter, req *http.Request)
...
+// 		default: genericError
+//			200: emptyRes
+func (o *Operation) deleteProfileHandler(rw http.ResponseWriter, req *http.Request) {
+	profileID := mux.Vars(req)["id"]
+
+	err := o.profileStore.DeleteProfile(profileID)
+
+	if err != nil {
+		if strings.Contains(err.Error(), storage.ErrValueNotFound.Error()) ||
+			strings.Contains(err.Error(), "Not Found: deleted") {
```
rolsonquadras, line 203: Is this specific to couchdb ?

rolsonquadras (approved):

pkg/restapi/verifier/operation/operations.go
grace0925: yes.. should i change corestore instead to ignore "Not Found:deleted"?

grace0925: @rolsonquadras

rolsonquadras: I would expect all store to be consistent with error messages ie - storage.ErrValueNotFound, rather than client handling individual store impl error messages. 

We can add this string check to couchdb and throw storage.ErrValueNotFound. This can be a separate PR - update edge-core and remove this check here.

DRK3: @grace0925 Also, when you rework the error handling per @rolsonquadras's comments, you should use `errors.Is(storage.ErrValueNotFound)`

pkg/restapi/verifier/operation/operations.go
```diff
@@ -182,6 +186,35 @@ func (o *Operation) getProfileHandler(rw http.ResponseWriter, req *http.Request)
...
+
+			return
+		}
+
+		commhttp.WriteErrorResponse(rw, http.StatusBadRequest, err.Error())
+
+		return
+	}
+
+	rw.WriteHeader(http.StatusOK)
```
DRK3, line 215: @grace0925 Note that http.StatusOK is automatically written if you don't write anything else, so you don't have to add it normally

PR #531, opened 2020-11-05
chore: update edv version

grace0925:
Signed-off-by: Grace Liu <grace.liu@securekey.com>

### DIFF
--- a/pkg/restapi/issuer/operation/operations.go
+++ b/pkg/restapi/issuer/operation/operations.go
@@ -558,7 +558,7 @@ func (o *Operation) createIssuerProfile(pr *ProfileRequest) (*vcprofile.DataProf
 
 	created := time.Now().UTC()
 
-	edvVaultID, err := o.createIssuerProfileVault()
+	edvVaultID, err := o.createIssuerProfileVault(didID)
 	if err != nil {
 		return nil, fmt.Errorf("fail to create issuer profile vault: %w", err)
 	}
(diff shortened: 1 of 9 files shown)


fqutishat (approved):

pkg/restapi/issuer/operation/operations.go
```diff
@@ -570,8 +570,12 @@ func (o *Operation) createIssuerProfile(pr *ProfileRequest) (*vcprofile.DataProf
 }
 
 // createIssuerProfileVault creates the vault associated with the profile
-func (o *Operation) createIssuerProfileVault() (string, error) {
-	vaultLocationURL, err := o.edvClient.CreateDataVault(&models.DataVaultConfiguration{ReferenceID: uuid.New().String()})
+func (o *Operation) createIssuerProfileVault(did string) (string, error) {
+	dataVaultConfig := &models.DataVaultConfiguration{Sequence: 0, Controller: did, ReferenceID: uuid.New().String(),
+		KEK:  models.IDTypePair{ID: "https://example.com/kms/12345", Type: "X25519KeyAgreementKey2019"},
```
rolsonquadras, line 575: `https://example.com/kms/12345` - Is this hardcoding expected ?

grace0925: it's just a hardcoded filler and the vault configuration is not used in edv right now.

rolsonquadras: Oh ok. If we need to fill the URI, then we can generate one dynamically using `uuid.New().URN()`.

rolsonquadras (approved):

PR #534, opened 2020-11-10
chore: update aries-framework-go revision

fqutishat:


pkg/internal/cryptosetup/cryptosetup_test.go
```diff
@@ -34,8 +34,11 @@ func TestPrepareJWECrypto(t *testing.T) {
 		keyHandleToBeCreated, err := keyset.NewHandle(mac.HMACSHA256Tag256KeyTemplate())
 		require.NoError(t, err)
 
+		testMACValue := []byte("testValue")
+		mockCrypto := crypto.Crypto{ComputeMACValue: testMACValue}
```
Baha-sk, line 38: you can use directly the default crypto implementation here:
`c, err := tinkcrypt.New()`

Baha-sk (approved):
LGTM, the unit tests doesn't need to use mock crypto with MAC key value for JWEEncrypter as it doesn't use MAC key

PR #584, opened 2021-02-04
feat: Saves documents into EDV

soluchok:
Request:
```curl
curl --location --request POST 'http://localhost:7878/vaults/did%3Akey%3Az6Mkp8naDFYfWioGsYQUP5abuDtNrcVgLBCRR6KWBfB8xtde%23z6Mkp8naDFYfWioGsYQUP5abuDtNrcVgLBCRR6KWBfB8xtde/docs' \
--header 'Content-Type: application/json' \
--data-raw '{
   "id":"M3aS9xwj8ybCwHkEiCJJR1",
   "content":{
      "name":"Andrii"
   }
}'
```

Response:
```
{
    "docID": "M3aS9xwj8ybCwHkEiCJJR1",
    "edvDocURI": "localhost:7777/encrypted-data-vaults/MTGyp2HCPrduHssnmGLw9i/documents/M3aS9xwj8ybCwHkEiCJJR1"
}
```

Signed-off-by: Andrii Soluk <isoluchok@gmail.com>


pkg/client/vault/client.go
```diff
@@ -42,9 +45,34 @@ type HTTPClient interface {
...
+}
+
+// Location consists of URI and zcap capability.
+type Location struct {
+	URI  string `json:"uri"`
+	ZCAP string `json:"zcap"`
+}
+
+// SavedDoc represents success response of SavedDoc function.
+type SavedDoc struct {
```
llorllale, line 67: This is modeled as "DocumentMetadata" in the spec

pkg/client/vault/client.go
```diff
@@ -205,3 +304,95 @@ func compressZCAP(zcap *zcapld.Capability) (string, error) {
...
+	jwe, err := encrypter.Encrypt(cek)
+	if err != nil {
+		return "", fmt.Errorf("encrypt: %w", err)
+	}
+
+	return jwe.FullSerialize(json.Marshal)
+}
+
+func encryptContent(content interface{}) ([]byte, []byte, error) {
+	cek := random.GetRandomBytes(crypto.DefKeySize)
```
llorllale, line 362: `crypto.DefKeySize` = 32, which means we're generating a random 256-bit key for encryption, which is fine, but I feel a little uneasy about importing the frameworks "default key size" constant. I suggest we define our own constant here... eg. `EncryptionKeySizeBytes`.

pkg/client/vault/client.go
```diff
@@ -205,3 +304,95 @@ func compressZCAP(zcap *zcapld.Capability) (string, error) {
...
+
+	jwe, err := encrypter.Encrypt(cek)
+	if err != nil {
+		return "", fmt.Errorf("encrypt: %w", err)
+	}
+
+	return jwe.FullSerialize(json.Marshal)
+}
+
+func encryptContent(content interface{}) ([]byte, []byte, error) {
```
llorllale, line 361: Calling in the cavalry: @Baha-sk

pkg/client/vault/client.go
```diff
@@ -205,3 +304,95 @@ func compressZCAP(zcap *zcapld.Capability) (string, error) {
...
+	cek := random.GetRandomBytes(crypto.DefKeySize)
+
+	kh, err := keyset.NewHandle(ecdh.NISTPECDHAES256GCMKeyTemplateWithCEK(cek))
```
llorllale, line 364: We shouldn't be doing key agreement here (ECDH)

pkg/client/vault/client.go
```diff
@@ -205,3 +304,95 @@ func compressZCAP(zcap *zcapld.Capability) (string, error) {
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("new ECDH encrypt: %w", err)
+	}
+
+	src, err := json.Marshal(content)
+	if err != nil {
+		return nil, nil, fmt.Errorf("marshal: %w", err)
+	}
+
+	encContent, err := enc.Encrypt(src, nil)
```
llorllale, line 384: Any reason why we are not using the framework's `crypto.webkms` package?

pkg/client/vault/client.go
llorllale: I think this function can be implemented as:

```go
	block, err := aes.NewCipher(cek)
	if err != nil {
		// error
	}
	nonce := make([]byte, 12)
	aesgcm, err := cipher.NewGCM(block)
	if err != nil {
		// error
	}
	encContent := aesgcm.Seal(nil, nonce, src, nil)
```

llorllale: ^ that would be using AESGCM, which is approved by NIST.

pkg/client/vault/client.go
```diff
@@ -205,3 +304,95 @@ func compressZCAP(zcap *zcapld.Capability) (string, error) {
...
+
+	pubKeyBytes, err := kmsImpl.ExportPubKeyBytes(kid)
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to retrieve main public key bytes from remote KMS: %w", err)
+	}
+
+	return pubKeyBytes, kid, nil
+}
+
+func encryptKey(wKMS KeyManager, wCrypto crypto.Crypto, cek []byte) (string, error) {
```
llorllale, line 329: You should be using `Crypto.WrapKey()` to encrypt the key.

llorllale: It seems the framework's `JWEEncrypt` does the wrapping for you... you just need to configure the remote `Crypto` implementation: https://github.com/hyperledger/aries-framework-go/blob/08fa4e3bce34a6a6798b80491f8ca4a651d83335/pkg/doc/jose/encrypter.go#L211-L215

pkg/client/vault/client.go
llorllale: So I believe you need two things:

* the webkms implementation of `KMS` (to create the keys and get the keyIDs back)
* the `JWEEncrypt` initialized with the webkms implementation of `Crypto` and the keyID/keyHandle

You'd just call `JWEEncrypt.Encrypt(plaintext []byte)` and it'll do the key wrapping for you.

pkg/client/vault/client.go
llorllale: disregard the above snippet... I suggest we use the framework's `JWEEncrypter`

pkg/client/vault/client.go
```diff
@@ -205,3 +304,95 @@ func compressZCAP(zcap *zcapld.Capability) (string, error) {
...
+	}
+
+	ecPubKey := &crypto.PublicKey{KID: jweCryptoKID}
+
+	err = json.Unmarshal(pubKeyBytes, ecPubKey)
+	if err != nil {
+		return "", fmt.Errorf("unmarshal: %w", err)
+	}
+
+	encrypter, err := jose.NewJWEEncrypt(jose.A256GCM, jose.DIDCommEncType, "", nil,
```
llorllale, line 347: `encType` should be "A256GCM" in this case:

* we are using a 32-byte key
* we are using AESGCM to encrypt

llorllale: You need to provide the `senderKID` and `senderKH` parameters because you are encrypting to your own key

pkg/client/vault/client.go
```diff
@@ -102,23 +136,73 @@ func (c *Client) CreateVault() (*CreatedVault, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("encrypt content: %w", err)
+	}
+
+	encKey, err := encryptKey(c.webKMS(vaultID, auth.KMS), c.webCrypto(vaultID, auth.KMS), key)
+	if err != nil {
+		return nil, fmt.Errorf("encrypt key: %w", err)
+	}
+
+	payload, err := sjson.Set(encKey, "content", base64.StdEncoding.EncodeToString(encContent))
```
Baha-sk, line 162: you don't need to go through the hassle of building the encrypted content and a shared secret separately here.

Use AFG directly to build the JWE envelope that contains the encrypted content and the wrapped cek all in one envelope.

I don't get why the encrypted key (wrapped cek)  need to be used as an index in `payload`.

cc @troyronda @DRK3

soluchok: @Baha-sk  you mean like I did here https://github.com/trustbloc/edge-service/pull/584/files#diff-286763af51a9a452468f5d071c9f2323ddac0dba0ce776c7ddc73ef9bad3662aR329

pkg/restapi/vault/operation/operations_test.go
```diff
@@ -101,6 +101,8 @@ func TestCreateVault(t *testing.T) {
 }
 
 func TestSaveDoc(t *testing.T) {
+	t.Skip()
```
DRK3, line 104: Is this supposed to be committed?

soluchok: No. PR is not ready, I have not added tests yet. I want to confirm the logic first.

Baha-sk:
`encryptContent()` and `encryptedKey()` are each building a JWE envelope. This is too much for saving 1 document.

If you want want to use the encrypted key as index, extracted directly from the JWE built for the content (it's called `encrypted_key` in the recipients header for multi recipients, or directly for single recipient JWE). There is no need to build the cek manually in the code, JWEEncrypter already does this internally.

pkg/client/vault/client.go
```diff
@@ -102,23 +133,66 @@ func (c *Client) CreateVault() (*CreatedVault, error) {
...
-type CreatedVault struct {
-	ID  string    `json:"id"`
-	EDV *Location `json:"edv"`
-	KMS *Location `json:"kms"`
+// SaveDoc saves a document by encrypting it and storing it in the vault.
+func (c *Client) SaveDoc(vaultID, id string, content interface{}, auth *Authorization) (*SavedDoc, error) {
+	const encryptionKeySizeBytes = 32
+
+	encContent, err := encryptContent(c.webKMS(vaultID, auth.KMS), c.webCrypto(vaultID, auth.KMS),
+		random.GetRandomBytes(encryptionKeySizeBytes))
```
Baha-sk, line 152: replace `random.GetRandomBytes(encryptionKeySizeBytes)` with `content` and remove line 149 above (`const encryptionKeySizeBytes = 32`)

pkg/client/vault/client.go
```diff
@@ -205,3 +294,56 @@ func compressZCAP(zcap *zcapld.Capability) (string, error) {
...
+
+	pubKeyBytes, err := kmsImpl.ExportPubKeyBytes(kid)
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to retrieve main public key bytes from remote KMS: %w", err)
+	}
+
+	return pubKeyBytes, kid, nil
+}
+
+func encryptContent(wKMS KeyManager, wCrypto crypto.Crypto, cek []byte) (string, error) {
```
Baha-sk, line 319: replace `cek` with `content` since we're encrypting content, not cek in this function

Baha-sk (changes requested):

pkg/client/vault/client.go
```diff
@@ -205,3 +290,61 @@ func compressZCAP(zcap *zcapld.Capability) (string, error) {
...
+	if err != nil {
+		return "", fmt.Errorf("create: %w", err)
+	}
+
+	pubKeyBytes, jweCryptoKID, err := prepareRemoteJWEKey(kidURL.(string), wKMS)
+	if err != nil {
+		return "", fmt.Errorf("prepare remote JWE key: %w", err)
+	}
+
+	ecPubKey := &crypto.PublicKey{KID: jweCryptoKID}
```
Baha-sk, line 331: you will need to populate all fields of ecPubKey, better yet unmarshall `pubKeyBytes` and use that instead because `kms.NISTP256ECDHKW`  type (and all of NISTP***ECDHKW and X25519 keys) is exported as `crypto.PublicKey`

pkg/client/vault/client_test.go
```diff
@@ -113,3 +159,31 @@ func TestCreateVault(t *testing.T) {
...
 	})
 }
+
+func TestSaveDoc(t *testing.T) {
+	const (
+		docID   = "id"
+		vaultID = "v_id"
+	)
+
+	t.Run("Unmarshal authorization arror", func(t *testing.T) {
```
Baha-sk, line 169: nitpick: typo `arror`

Baha-sk (approved):

pkg/client/vault/client.go
```diff
@@ -205,3 +343,59 @@ func compressZCAP(zcap *zcapld.Capability) (string, error) {
...
+	return pubKeyBytes, nil
+}
+
+func encryptContent(wKMS KeyManager, wCrypto crypto.Crypto, content interface{}) (string, error) {
+	src, err := json.Marshal(content)
+	if err != nil {
+		return "", fmt.Errorf("marshal: %w", err)
+	}
+
+	_, kidURL, err := wKMS.Create(kms.NISTP256ECDHKW)
```
llorllale, line 372: There should be no need to create an asymmetric key for encryption. In this case it results in additional network overhead. The encryption key should be created locally.

llorllale (approved):

rolsonquadras (approved):

PR #600, opened 2021-02-10
fix: Vault server - generates id when it is empty.

soluchok:
Signed-off-by: Andrii Soluk <isoluchok@gmail.com>

### DIFF
--- a/pkg/restapi/vault/operation/operations.go
+++ b/pkg/restapi/vault/operation/operations.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/trustbloc/edge-core/pkg/log"
+	"github.com/trustbloc/edv/pkg/edvutils"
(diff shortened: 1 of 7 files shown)


rolsonquadras (approved):

pkg/restapi/vault/operation/operations.go
```diff
@@ -37,12 +38,16 @@ var logger = log.New("vault-operation")
 
 // Operation defines handlers for vault service.
 type Operation struct {
-	vault vault.Vault
+	vault      vault.Vault
+	GenerateID func() (string, error)
```
llorllale, line 42: nit: it's weird to have this exposed

llorllale (approved):

pkg/restapi/vault/operation/operations.go
soluchok: Did it only for testing purposes.

PR #607, opened 2021-02-16
feat: confidential-storage-hub /compare

llorllale:


pkg/restapi/csh/operation/openapi/caveat.go
```diff
@@ -0,0 +1,102 @@
+// Code generated by go-swagger; DO NOT EDIT.
```
fqutishat, line 1: where is the command to generate these files in make file

llorllale: Not implemented. Need to convert OpenAPI v3 spec to Swagger 2.0. Even then, have to manually fix 2 lines in the generated code so as not to break the `check-openapi-specs` target (seems like a bug in goswagger, maybe).

In short, I generated the swagger models using the goswagger CLI in order to avoid having to manually code the boilerplate code, which is significant.

fqutishat (approved):

rolsonquadras (approved):

PR #612, opened 2021-02-18
test: confidential-storage-hub: /compare BDD tests

llorllale:
* added BDD tests for /compare
* converted OpenAPI spec to swagger 2.0 because go-swagger does not support OpenAPI 3
* fix: payload of EDV document JWEs were assumed to be generic JSON documents instead of EDV StructuredDocuments
* fix: ineffectual setting of HTTP response headers
* fix: missing HTTP client configuration on EDV client
* fix: keystore URL determination with a temp workaround
* fix: missing action arg for zcapld.NewHTTPSigner()


test/bdd/pkg/chs/openapi/client/confidential_storage_hub_client.go
```diff
@@ -0,0 +1,111 @@
+// Code generated by go-swagger; DO NOT EDIT.
```
fqutishat, line 1: Please move it under pkg/client https://github.com/trustbloc/edge-service/tree/main/pkg/client/csh

fqutishat: other components like comparator need to use the client

fqutishat (approved):

rolsonquadras (approved):

PR #635, opened 2021-02-25
fix: vault-server, comparator

llorllale:
* vault-server: the EDV StructuredDocument model was not being used to  save documents
* comparator: added DocAttrPath so that the client can authorize comparisons against specific portions of the document. TODO encode this path into the zcap token.


llorllale:
@rolsonquadras @fqutishat 

`docAttrPath` was added to Comparator's `Authorization.scope`

fqutishat (approved):

soluchok (approved):

PR #647, opened 2021-03-03
feat: add deriveCredential for holder

fqutishat:

### DIFF
--- a/pkg/restapi/holder/operation/models.go
+++ b/pkg/restapi/holder/operation/models.go
@@ -43,3 +43,23 @@ type SignPresentationOptions struct {
 	Challenge          string     `json:"challenge,omitempty"`
 	Domain             string     `json:"domain,omitempty"`
 }
+
+// DeriveCredentialRequest is request for deriving credential.
+type DeriveCredentialRequest struct {
+	// Raw Credential from which a new credential will be derived
+	Credential json.RawMessage `json:"verifiableCredential,omitempty"`
+	// Frame is JSON-LD frame used for selective disclosure.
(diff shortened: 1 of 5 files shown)


pkg/restapi/holder/operation/models.go
```diff
@@ -43,3 +43,16 @@ type SignPresentationOptions struct {
 	Challenge          string     `json:"challenge,omitempty"`
 	Domain             string     `json:"domain,omitempty"`
 }
+
+// DeriveCredentialRequest is request for deriving credential.
+type DeriveCredentialRequest struct {
```
fqutishat, line 48: @troyronda @sudeshrshetty derive request doesn't have nonce field in vc-http-api https://github.com/w3c-ccg/vc-http-api/commit/a8cb2d0c370de4dfbb6b8437d84bb63b9880f5c1#diff-74dfbc770d6e21cb974f297ba8f1bc7015ece556537a6202e87b78622728a0cfR562

sudeshrshetty: This is optional

sudeshrshetty (approved):

pkg/restapi/holder/operation/models.go
troyronda: I think we should add an option for nonce, so that we have an HTTP API that can handle the full derivation scope.

fqutishat: @troyronda done. added option for nonce

fqutishat: https://github.com/w3c-ccg/vc-http-api/issues/136

troyronda (approved):

PR #656, opened 2021-03-05
fix: generates a random nonce if not provided for deriveCredentials

fqutishat:


sudeshrshetty (approved):

pkg/restapi/holder/operation/operations_test.go
```diff
@@ -383,7 +383,7 @@ func TestDeriveCredentials(t *testing.T) {
 		require.NotEmpty(t, derived)
 		require.Len(t, derived.Proofs, 1)
 		require.Equal(t, derived.Proofs[0]["type"], "BbsBlsSignatureProof2020")
-		require.Empty(t, derived.Proofs[0]["nonce"])
+		require.NotEmpty(t, derived.Proofs[0]["nonce"])
```
troyronda, line 386: This feels like we had the correct behavior previously - if nonce is actually optional.

PR #664, opened 2021-03-08
feat: handle nonce cases

fqutishat:
- Case 1: No option provided - generate nonce.
- Case 2: Option provided with empty string - use empty.
- Case 3: option provided with nonce - use nonce from option.


rolsonquadras (approved):

troyronda:
We will have to see if Case 1 goes away or stays.

troyronda (approved):

PR #671, opened 2021-03-14
fix: add vc status list for each profile

fqutishat:

### DIFF
--- a/pkg/doc/vc/status/csl/csl.go
+++ b/pkg/doc/vc/status/csl/csl.go
@@ -23,7 +23,8 @@ import (
 )
 
 const (
-	vcContext = "https://www.w3.org/2018/credentials/v1"
+	vcContext               = "https://www.w3.org/2018/credentials/v1"
+	jsonWebSignature2020Ctx = "https://w3c-ccg.github.io/lds-jws2020/contexts/lds-jws2020-v1.json"
 	// Context for Revocation List 2020
 	Context = "https://w3id.org/vc-revocation-list-2020/v1"
 	// CredentialStatusType credential status type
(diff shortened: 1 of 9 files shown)


sudeshrshetty (approved):

pkg/doc/vc/status/csl/csl.go
```diff
@@ -23,7 +23,8 @@ import (
 )
 
 const (
-	vcContext = "https://www.w3.org/2018/credentials/v1"
+	vcContext               = "https://www.w3.org/2018/credentials/v1"
+	jsonWebSignature2020Ctx = "https://w3c-ccg.github.io/lds-jws2020/contexts/lds-jws2020-v1.json"
```
troyronda, line 27: this isn't the canonical context

it's: https://w3id.org/security/jws/v1

pkg/doc/vc/status/csl/csl.go
```diff
@@ -296,6 +297,11 @@ func (c *CredentialStatusManager) createVC(vcID string,
 	profile *vcprofile.DataProfile) (*verifiable.Credential, error) {
 	credential := &verifiable.Credential{}
 	credential.Context = []string{vcContext, Context}
+
+	if profile.SignatureType == vccrypto.JSONWebSignature2020 {
```
troyronda, line 301: note: BBS would also need another context if that key is used.

troyronda (approved):

pkg/doc/vc/status/csl/csl.go
fqutishat: done. change it to ttps://w3id.org/security/jws/v1

troyronda: Looks like it got reverted.

Issue #735, opened 2022-07-12
Optimize MongoDB usage


DRK3:
Currently VCS uses the Aries storage interface from [Aries-Framework-Go](https://github.com/hyperledger/aries-framework-go/tree/main/spi/storage) with implementations from [Aries-Framework-Go-Ext](https://github.com/hyperledger/aries-framework-go-ext/tree/main/component/storage) (and also an in-memory implementation in [Aries-Framework-Go](https://github.com/hyperledger/aries-framework-go/tree/main/component/storageutil/mem))

This storage interface was designed around the limitations of the [Encrypted Data Vault spec](https://github.com/decentralized-identity/edv-spec), to allow an EDV storage provider implementation (like [ours](https://github.com/hyperledger/aries-framework-go/tree/main/component/storage/edv)) to be swapped out with a non-EDV storage implementation without having to make code changes.

This approach is introduces limitations for non-EDV storage, but for Aries-Framework-Go since it wants to have the option of both encrypted and non-encrypted storage. In VCS this isn't as useful since it doesn't use EDV anymore (as of [this PR](https://github.com/trustbloc/vcs/pull/745)).

Two key limitations of the Aries interface are:
1. Storage overhead (in terms of storage space required) when storing JSON due to the need to "wrap" data and tags in a way that ensures that there won't be name conflicts and ensures that data querying works in the same way as EDV.
2. Inflexibility with querying for JSON data and potential tag overhead. With the Aries storage interface, you can't query directly on the JSON - instead, you must define "tags" when storing the data that you can then query on later. Sometimes these tags are data that is already contained within the JSON. For that data to be queryable, you have to add tags, even though it's already in the JSON. The result is that you get data duplication and also you don't get the flexibility to freely query your JSON later on without making code/database changes. This lack of flexibility and data duplication is a limitation of EDV due to how its encrypted nature, but it gets carried over into our other storage providers in order to ensure consistency and swapability among implementations

I've already added functionality to our MongoDB storage implementation in Aries-Framework-Go-Ext to allow us to bypass these restrictions, but to use it in VCS requires some changes to how we do storage in VCS.

PR #758, opened 2022-08-04
feat: Storage improvements

DRK3:
Made various improvements to how VCS handles data storage, including:

* Reduced storage space used in MongoDB by removing unnecessary data wrapping and data duplication
* Improvements to database and ID names for all database providers
* Improvements to storage models
* Refactored storage code into separate packages
* Master key now stored directly as bytes without base64 encoding

cllses #735


fqutishat (approved):

pkg/storage/holderprofile.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (h *HolderProfileStore) Put(profile HolderProfile) error {
+	mongoDBStore, isMongoDBStore := h.ariesStore.(*mongodb.Store)
+	if isMongoDBStore {
+		// Save space in the database by using the profile name as the MongoDB document _id field
+		// and removing the profile name from the profile JSON.
+		id := profile.Name
+		profile.Name = ""
+
+		return mongoDBStore.PutAsJSON(id, profile)
```
rolsonquadras, line 46: Looks like we are doing MongoDB specific conversion in all the functtion.

rolsonquadras (changes requested):
Looks like we are doing MongoDB specific conversion in all the function and it beats the purpose of having a common interface. Probably we need to relook at the use of common interface and requirements of the projects, ie, whether we really need multiple DB vs one DB  - let's discuss this. For now, adding change request to avoid merge.

pkg/doc/vc/crypto/crypto.go
```diff
@@ -188,7 +189,7 @@ type Crypto struct {
 }
 
 // SignCredential signs vc.
-func (c *Crypto) SignCredential(dataProfile *vcprofile.DataProfile, vc *verifiable.Credential,
+func (c *Crypto) SignCredential(dataProfile *vcsstorage.DataProfile, vc *verifiable.Credential,
```
rolsonquadras, line 192: Do we need to put everything in a storage package? Can't we modularize storage for each domain the way it is? Like issuer, verifier, holder, csl, crypto?

aholovko: I was looking into this as well. Can we minimize changes here and make refactoring the next task as part of implementing new API (we have a better context on how it is going to be used). cc @DRK3, @rolsonquadras

DRK3: @rolsonquadras I could move the interface and data type definitions into separate packages, or perhaps move them into their respective operations files. (Part of the reason for moving the data type/models into the storage folder was to avoid circular dependencies with the original way I implemented it in my original iteration of the PR, but now that's not an issue)

However, on a similar note as what @aholovko said, I think it would be more productive to not rework this PR too many times considering the large refactor work that @aholovko and @vkubiv are working on. I'd suggest merging this in as-is and then I'll work with @aholovko and @vkubiv to refactor the storage code in a better way that also works with the new design. The most important part of this PR is to get the MongoDB storage optimizations in.

aholovko (approved):
with future improvements in mind :)

rolsonquadras (approved):
The PR contains MongoDB optimization changes along with significant package structure changes. @aholovko will come up with a proposal for the whole project structure, including the storage layer - approving the PR with this caveat.

@fqutishat @sudeshrshetty @troyronda thoughts ?

fqutishat (approved):

PR #762, opened 2022-08-05
fix: BDD tests tags not working

DRK3:


DRK3:
BDD test tags are still working (the env vars get picked up), no changes needed (except to remove the check_integration.sh script as it's not in use anymore) @aholovko @vkubiv

PR #772, opened 2022-08-25
feat: verifier profile management REST API + stubs for issuer API

aholovko:
Signed-off-by: Andrii Holovko <andriy.holovko@gmail.com>


Makefile
```diff
@@ -34,13 +34,14 @@ all: checks unit-test bdd-test
 .PHONY: checks
 checks: license lint open-api-spec
 
-.PHONY: mocks
-mocks:
+.PHONY: generate
+generate:
 	@GOBIN=$(GOBIN_PATH) go install github.com/golang/mock/mockgen@$(MOCK_VERSION)
+	@GOBIN=$(GOBIN_PATH) go install github.com/deepmap/oapi-codegen/cmd/oapi-codegen@latest
 	@go generate ./...
```
sudeshrshetty, line 41: why we have common generate command for mocks & open apis here? they are dependent on each other?

sudeshrshetty (approved):

fqutishat (approved):

Makefile
aholovko: "generate" is a global command, we can't differentiate one type of "generate" from another
in our case, they come hand-in-hand, ie.
```
//go:generate oapi-codegen --config=openapi.cfg.yaml ../../../../api/openapi.yaml
//go:generate mockgen -destination controller_mocks_test.go ...
```

sudeshrshetty: if some one wants to generate open api spec, why he has to wait for mocks to get generated?

I am fine with having common `generate` command which generates everything (grouping other sub generate commands), but I would suggest having separate for mock, open api...and common generate command as umbrella command

aholovko: makes sense, I will try to come up with some solution in the upcoming PRs cc@sudeshrshetty

PR #776, opened 2022-09-01
feat: issuer profile controller, validators, and general error handling.

vkubiv:


test/bdd/fixtures/docker-compose.yml
```diff
@@ -20,6 +20,7 @@ services:
       - KMSSECRETS_DATABASE_TYPE=mongodb
       - KMSSECRETS_DATABASE_URL=mongodb://mongodb.example.com:27017
       - KMSSECRETS_DATABASE_PREFIX=issuer_
+      - VC_SECRET_LOCK_KEY_PATH=/etc/tls/secret-lock.key
```
rolsonquadras, line 23: How is key rotation handled for KMS secret lock ?

pkg/restapi/v1/util/auth.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/labstack/echo/v4"
+
+	"github.com/trustbloc/vcs/pkg/restapi/resterr"
+)
+
+func GetOrgIDFromOIDC(ctx echo.Context) (string, error) {
```
rolsonquadras, line 18: I don't see other code in this PR calling this function.

vkubiv: Maybe controller.go file is collapsed in the review section by default, so the "find" command does not work. I used it here https://github.com/trustbloc/vcs/pull/776/files#diff-227f4ba89c786112771c32162dc814dece16c31c38c131df102921d58f240619R90

test/bdd/fixtures/docker-compose.yml
vkubiv: Not handled now, the same problem as with key rotation in KMS. We can implement rotation, but I think it better to create a separate issue, because it is not trivial

rolsonquadras: Yes, let's create an issue - this is an important topic.

vkubiv: https://github.com/trustbloc/vcs/issues/783

aholovko (approved):

PR #799, opened 2022-09-20
feat: oidc tx manager.

vkubiv:

### DIFF
--- a/pkg/storage/redis/oidctxstore/tx_store.go
+++ b/pkg/storage/redis/oidctxstore/tx_store.go
@@ -1,0 +1,64 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package oidctxstore
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/go-redis/redis/v8"
+)
+
+// TxStore stores oidc transactions in redis.
+type TxStore struct {
+	client    *redis.Client
+	opTimeout time.Duration
+}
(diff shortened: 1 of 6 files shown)


pkg/storage/redis/oidctxstore/tx_store_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+func createRedisClient() *redis.Client {
+	return redis.NewClient(&redis.Options{
+		Addr:     redisDBHost,
+		Password: "", // no password set
+		DB:       0,
+	})
+}
+
+func pingMongoDB() error {
```
aholovko, line 166: fix to `pingRedisDB`

aholovko (approved):

pkg/storage/redis/oidctxstore/tx_store.go
```diff
@@ -0,0 +1,64 @@
...
+// New creates TxStore.
+func New(client *redis.Client, opTimeout time.Duration) *TxStore {
+	return &TxStore{
+		client:    client,
+		opTimeout: opTimeout,
+	}
+}
+
+// GetAndDelete get and then delete transaction by one time token.
+func (ts *TxStore) GetAndDelete(key string) ([]byte, bool, error) {
```
sudeshrshetty, line 33: what is the use of this function?

vkubiv: Atomically get transaction by one-time token. after this, it will be deleted from the store. The method is used in the end stage of interaction. To mitigate replay attacks.

sudeshrshetty (approved):

PR #810, opened 2022-09-26
feat: add oidc4vp support to create verifier profile.

vkubiv:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -62,6 +62,7 @@ require (
 	github.com/go-openapi/jsonpointer v0.19.5 // indirect
 	github.com/go-openapi/swag v0.21.1 // indirect
 	github.com/go-sql-driver/mysql v1.6.0 // indirect
+	github.com/go-test/deep v1.0.8 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b // indirect
 	github.com/golang/protobuf v1.5.2 // indirect
--- a/pkg/restapi/v1/verifier/controller_test.go
+++ b/pkg/restapi/v1/verifier/controller_test.go
@@ -9,19 +9,28 @@ package verifier
 import (
 	"bytes"
 	_ "embed"
+	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
 	"testing"
 
 	"github.com/golang/mock/gomock"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/presexch"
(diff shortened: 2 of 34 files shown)


go.mod
```diff
@@ -62,6 +62,7 @@ require (
 	github.com/go-openapi/jsonpointer v0.19.5 // indirect
 	github.com/go-openapi/swag v0.21.1 // indirect
 	github.com/go-sql-driver/mysql v1.6.0 // indirect
+	github.com/go-test/deep v1.0.8 // indirect
```
mishasizov-SK, line 65: Do we have any company restrictions to import 3rd party packages?

vkubiv: It is used just for tests. So should be any proplems

pkg/restapi/v1/verifier/controller_test.go
```diff
@@ -43,22 +52,33 @@ var (
...
+	kms.ECDSAP384TypeDER,
+	kms.RSAPS256Type,
+	kms.BLS12381G2Type,
+}
+
 //nolint:gochecknoglobals
 var (
 	verificationChecks = &verifiersvc.VerificationChecks{
-		Credential: &verifiersvc.CredentialChecks{
+		Credential: verifiersvc.CredentialChecks{
```
aholovko, line 69: is there a specific reason why Credential is value type while Presentation below is reference?

pkg/restapi/v1/verifier/openapi.gen.go
```diff
@@ -9,42 +9,53 @@ import (
...
 	// URI of the verifier.
 	Url *string `json:"url,omitempty"`
 }
 
+// Model for OIDC4VPConfig configuration.
+type OIDC4VPConfig struct {
+	// DID method of the DID to be used for signing.
+	DidMethod externalRef0.DIDMethod `json:"didMethod"`
+
+	// Cryptographic algorithms used to sign oipd vp request object.
```
aholovko, line 42: oipd -> oidc?

do we need `Ro` prefix in `RoSigningAlgorithm`? if this is important to emphasize "request object" maybe name it something like "RequestSigningAlgorithm" ("ro" has a collocation with "read-only")

aholovko (approved):

pkg/restapi/v1/verifier/controller_test.go
vkubiv: Credential is not optional, Presentation is optional

pkg/restapi/v1/verifier/openapi.gen.go
vkubiv: "Request" is confusing, but yes "ro" is definitely not the best name. 

I will include fix for oipd and "ro" in next PR

mishasizov-SK (approved):

PR #841, opened 2022-10-05
feat: oidc4vp check authorizaton response.

vkubiv:


pkg/restapi/v1/verifier/controller.go
```diff
@@ -35,8 +39,30 @@ import (
...
+type authorizationResponse struct {
+	IDToken string
+	VPToken string
+	State   string
+}
+
+type IDTokenClaims struct {
+	VPToken struct {
+		PresentationSubmission *presexch.PresentationSubmission `json:"presentation_submission"`
+	} `json:"_vp_token"`
```
mishasizov-SK, line 55: Please consider to extract it so the separate struct.

mishasizov-SK (approved):

PR #847, opened 2022-10-07
feat: add wellknown api

skynet2:
closes: https://github.com/trustbloc/vcs/issues/839


fqutishat (approved):

pkg/restapi/v1/devapi/controller.go
```diff
@@ -0,0 +1,70 @@
...
+	if scheme := ctx.Request().URL.Scheme; len(scheme) > 0 {
+		contextURL.WriteString(scheme + "://")
+	} else {
+		contextURL.WriteString("https://")
+	}
+
+	if hostName := ctx.Request().URL.Hostname(); len(hostName) > 0 {
+		contextURL.WriteString(hostName + "/")
+	} else {
+		contextURL.WriteString("localhost")
```
aholovko, line 61: this looks strange in production code, plz elaborate on this logic

pkg/service/didconfiguration/didconfiguration_service.go
```diff
@@ -0,0 +1,200 @@
...
+		if err != nil {
+			return nil, resterr.NewValidationError(resterr.SystemError, "profileID",
+				err)
+		}
+		cred.Issuer = verifiable.Issuer{
+			ID: profile.SigningDID.DID,
+		}
+		cred.Subject = map[string]interface{}{
+			"id":     profile.SigningDID.DID, // todo nothing in JSON ???
+			"origin": fmt.Sprintf("%s://%s", u.Scheme, u.Hostname()),
```
aholovko, line 117: shouldn't this be an URL from profile?

test/bdd/pkg/vc-devapi/vc_devapi_steps.go
```diff
@@ -0,0 +1,86 @@
...
+func (s *Steps) httpGet(profileType string, id string, credType string) error {
+	s.credType = credType
+
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet,
+		fmt.Sprintf("http://localhost:8075/%s/profiles/%s/well-known/did-config",
+			profileType, id), http.NoBody)
+
+	req.Header.Add("X-API-Key", "rw_token")
+
+	if err != nil {
```
aholovko, line 60: move error checking before `req.Header.Add`

test/bdd/pkg/vc-devapi/vc_devapi_steps.go
```diff
@@ -0,0 +1,86 @@
...
+	return &Steps{bddContext: ctx}
+}
+
+// RegisterSteps registers VC scenario steps.
+func (s *Steps) RegisterSteps(sc *godog.ScenarioContext) {
+	sc.Step(`^I request did config for "([^"]*)" with ID "([^"]*)" and type "([^"]*)"$`, s.httpGet)
+	sc.Step(`^I receive response with status code "([^"]*)" for didconfig$`, s.checkResponseStatus)
+}
+
+func (s *Steps) checkResponseStatus(status string) error {
```
aholovko, line 36: can we perform some basic checks for didconfig from response as well? like some properties that for sure should be there...

aholovko (changes requested):

pkg/restapi/v1/devapi/controller.go
fqutishat: I agree with that

skynet2: We need to build absolute URI. in HTTP request, standard Request().URL there is no possibility of receiving full URL (RequestURI() in some cases can return only Path, without schema and domain, so I combine URL by myself from request, that code basically can be moved to some helper. cc @aholovko @fqutishat

aholovko: Stas, you are using that URL as a value of @context - I think it's not correct, there should be constant value "https://identity.foundation/.well-known/did-configuration/v1" (JSON-LD context for "DID Configuration" document). Probably, you confused that with https://identity.foundation/.well-known/resources/did-configuration/#did-configuration-uri that describes where that document should be hosted. cc @skynet2

test/bdd/pkg/vc-devapi/vc_devapi_steps.go
skynet2: fixed

pkg/restapi/v1/devapi/controller.go
skynet2: you are right @aholovko , fixed

PR #870, opened 2022-10-14
feat: update additional structured logs

Baha-sk:


internal/pkg/log/fields.go
```diff
@@ -18,7 +18,9 @@ const (
 	FieldUserLogLevel        = "userLogLevel"
 	FieldID                  = "id"
 	FieldName                = "name"
+	FieldOut                 = "out"
```
bstasyszyn, line 21: I think this should be more descriptive. What is actually being logged here?

Baha-sk: It's mainly for docker-compose command  in bddtests-test.go, in this PR it's at line 130 right now.

bstasyszyn: So, maybe WithDockerResponse or if you want something more generic, WithCommandResponse?

Baha-sk: i'll rename to command instead

bstasyszyn (approved):

internal/pkg/log/fields.go
Baha-sk: done

PR #890, opened 2022-10-18
feat: Credential Status API - update vc status

mishasizov-SK:
Signed-off-by: Mykhailo Sizov <mykhailo.sizov@securekey.com>

### DIFF
--- a/pkg/doc/vc/crypto/crypto.go
+++ b/pkg/doc/vc/crypto/crypto.go
@@ -150,8 +150,20 @@ type Crypto struct {
 	documentLoader ld.DocumentLoader
 }
 
-// SignCredentialLDP adds verifiable.LinkedDataProofContext to the vc.
-func (c *Crypto) SignCredentialLDP(
+func (c *Crypto) SignCredential(
+	signerData *vc.Signer, vc *verifiable.Credential, opts ...SigningOpts) (*verifiable.Credential, error) {
+	switch signerData.Format {
+	case vcsverifiable.Jwt:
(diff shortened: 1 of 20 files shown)


pkg/doc/vc/crypto/crypto.go
```diff
@@ -150,8 +150,20 @@ type Crypto struct {
 	documentLoader ld.DocumentLoader
 }
 
-// SignCredentialLDP adds verifiable.LinkedDataProofContext to the vc.
-func (c *Crypto) SignCredentialLDP(
+func (c *Crypto) SignCredential(
```
mishasizov-SK, line 153: created a generic method instead of `SignCredentialLDP` and `SignCredentialJWT`

pkg/service/verifycredential/revocation/revocation_service.go
```diff
@@ -114,5 +115,5 @@ func (s *Service) sendHTTPRequest(req *http.Request, status int, token string) (
 		return nil, fmt.Errorf("failed to read response body for status %d: %s", resp.StatusCode, string(body))
 	}
 
-	return body, nil
+	return bytes.Trim(body, "\n"), nil
```
fqutishat, line 118: why we need trim here

mishasizov-SK: ctx.JSON(), ctx.String() and similar methods from echo adds `/n` to the end of payload body. 
For LDP VC it's ok, but for jwt it makes problems in decoding.

fqutishat (approved):

aholovko (approved):

PR #912, opened 2022-10-25
feat: vc authorize flow

skynet2:


docs/v1/openapi.yaml
```diff
@@ -503,6 +522,29 @@ paths:
                 - code
                 - code_verifier
         description: ''
+  /oidc/callback:
+    get:
+      summary: OIDC Callback
+      tags:
+        - oidc4vc
+      operationId: oidc-callback
+      description: OIDC callback for storing external issuer auth code and continue our oidc auth flow.
```
aholovko, line 531: ... for handling response from issuer's OIDC provider and continue our OIDC authorize flow.

pkg/restapi/v1/oidc4vc/controller.go
```diff
@@ -125,12 +150,22 @@ func (c *Controller) GetOidcAuthorize(e echo.Context, params GetOidcAuthorizePar
...
+		return apiUtil.WriteOutput(e)(nil, errors.New("op_state is required"))
+	}
+
 	ar, err := c.oauth2Provider.NewAuthorizeRequest(ctx, req)
 	if err != nil {
 		return resterr.NewFositeError(resterr.FositeAuthorizeError, e, c.oauth2Provider, err).WithAuthorizeRequester(ar)
 	}
 
-	// TODO: login and get consent from the user
+	mapped, mapOk := ar.(*fosite.AuthorizeRequest)
```
aholovko, line 162: is it possible that ar will not be of type `*fosite.AuthorizeRequest`, can this be replaced with `ar.(*fosite.AuthorizeRequest).State = *params.OpState`?

skynet2: I think no, but it's an external library, so we can not be 100% sure. it safer to properly handle that, rather than have panic

docs/v1/openapi.yaml
skynet2: done

aholovko (approved):

fqutishat (approved):

PR #935, opened 2022-11-04
chore: enable metrics http handler

Baha-sk:


bstasyszyn (approved):

Baha-sk:
it seems this doesn't register the echo handler, only the prometheus one, I need to wrap it into an echo handler instead (since we don't have a concret http.Server at all)

fqutishat (approved):

PR #938, opened 2022-11-07
feat: preauth endpoints

skynet2:
Signed-off-by: Stas D <stanislav.dmytryshyn@securekey.com>


fqutishat (approved):

cmd/vc-rest/startcmd/start.go
```diff
@@ -299,6 +297,14 @@ func buildEchoHandler(conf *Configuration, cmd *cobra.Command) (*echo.Echo, erro
 		StateStore:              oidc4StateStore,
 		IssuerInteractionClient: issuerInteractionClient,
 		IssuerVCSPublicHost:     conf.StartupParameters.hostURLExternal,
+		DefaultHTTPClient:       httpClient,
+		PreAuthorizeClient: func() *http.Client {
+			client := getDefaultHttpClient(tlsConfig)
+			client.CheckRedirect = func(req *http.Request, via []*http.Request) error {
```
aholovko, line 303: do we really want to disable following redirects? plz elaborate on this...

one more issue - it's not clear what is the purpose of `getDefaultHttpClient` function? It returns client by reference so any changes to this instance will affect it in other places as well.

pkg/oauth2client/oauth_client.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+type Client struct {
+}
+
+func NewOAuth2Client() *Client {
+	return &Client{}
+}
+
+func (c *Client) Exchange(
```
aholovko, line 25: there is no usage of this method, is it necessary?

pkg/oauth2client/oauth_client.go
```diff
@@ -0,0 +1,51 @@
...
+	ctx context.Context,
+	cfg oauth2.Config,
+	code string,
+	opts ...oauth2.AuthCodeOption,
+) (*oauth2.Token, error) {
+	return (&cfg).Exchange(ctx, code, opts...)
+}
+
+func (c *Client) ExchangeWithCustomClient(
```
aholovko, line 34: consider moving `client *http.Client` to the method above and just have a single `Exchange` func that accepts `client *http.Client` - for now there is no need for that versatility

aholovko (changes requested):
consider splitting this PR into 2: one for pre-auth functionality and another one for customizing your factory with http client

cmd/vc-rest/startcmd/start.go
skynet2: getDefaultHttpClient  - always returns a new reference to a new client. 
we need to disable redirect, as we don't need it for auto pre-auth flow https://github.com/trustbloc/vcs/pull/938/files#diff-6e396ac797089209a3155761638fd798eefc3e9c1d780e9019d8793e33e07371R409

pkg/oauth2client/oauth_client.go
skynet2: Done @aholovko

aholovko (approved):

PR #950, opened 2022-11-11
feat: prepare and sign credential for credential API

aholovko:
Signed-off-by: Andrii Holovko <andriy.holovko@gmail.com>


skynet2 (approved):

pkg/service/oidc4vc/oidc4vc_service.go
```diff
@@ -232,11 +262,51 @@ func (s *Service) PrepareCredential(
...
+			ID:           req.DID,
+			CustomFields: claimData,
+		},
+		Issued: util.NewTime(time.Now()),
+	}
+
+	var credential interface{}
+
+	switch tx.CredentialFormat {
+	case vcsverifiable.Jwt:
```
mishasizov-SK, line 287: Is it possible to use crypto service to avoid switch-case statement? 
eventually it does he same 
https://github.com/trustbloc/vcs/blob/main/pkg/doc/vc/crypto/crypto.go#L153

pkg/service/oidc4vc/oidc4vc_service.go
```diff
@@ -232,11 +262,51 @@ func (s *Service) PrepareCredential(
...
+		claims, jwtClaimsErr := vc.JWTClaims(false)
+		if jwtClaimsErr != nil {
+			return nil, fmt.Errorf("create jwt claims: %w", jwtClaimsErr)
+		}
+
+		credential, err = claims.MarshalUnsecuredJWT()
```
mishasizov-SK, line 293: Dont we need to call `claims.MarshalJWS` here?

pkg/service/oidc4vc/oidc4vc_service.go
aholovko: there is no signer in our disposition, we just prepare data in a format understandable by issuer's endpoint `/issuer/profiles/{profileID}/credentials/issue` and delegate signing to it, in essence, `credentialService` here is a client to that endpoint

pkg/service/oidc4vc/oidc4vc_service.go
aholovko: see above

pkg/service/oidc4vc/oidc4vc_service.go
mishasizov-SK: ok, thanks for explanation

mishasizov-SK (approved):

PR #1057, opened 2023-01-30
feat: add support for credential offer uri

skynet2:
Signed-off-by: Stas D <stanislav.dmytryshyn@securekey.com>

### DIFF
--- a/cmd/vc-rest/startcmd/params.go
+++ b/cmd/vc-rest/startcmd/params.go
@@ -198,6 +198,14 @@ const (
 	requestObjectRepositoryS3RegionEnvKey    = "REQUEST_OBJECT_REPOSITORY_S3_REGION"
 	requestObjectRepositoryS3RegionFlagUsage = "request-object S3 Region"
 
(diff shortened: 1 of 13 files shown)


skynet2:
hi @sudeshrshetty @fqutishat , as per last conversation with @sudeshrshetty we will use "credential_offer_uri" as default implementation for initiate issuance response. 
That means that we need to store initiate issuance response to the external server ( in our case s3 ).
For the current implementation, I made it as optional step [credential-offer-repository-s3-bucket](https://github.com/trustbloc/vcs/pull/1057/files#diff-14dd31deba2e8f73a5e26eb8cb52b9faa04399a52a32ad084a8b39b2d48642f7R494) so we will put data, only if s3 is properly configured with env variables **CREDENTIAL_OFFER_REPOSITORY_S3_BUCKET** and **CREDENTIAL_OFFER_REPOSITORY_S3_REGION**.

@fqutishat I think it also make sense to add some s3 mock to our bdd tests, I will create an issue for that so we will not forget about that. currently bdd tests will not cover **credential_offer_uri** scenario

fqutishat (approved):

PR #1194, opened 2023-03-14
perf: vp flow

skynet2:


pkg/service/verifypresentation/verifypresentation_service.go
```diff
@@ -58,13 +61,20 @@ func New(config *Config) *Service {
...
 
+var logger = log.New("verify-presentation")
+
 func (s *Service) VerifyPresentation(
 	presentation *verifiable.Presentation,
 	opts *Options,
 	profile *profileapi.Verifier) ([]PresentationVerificationCheckResult, error) {
+	startTime := time.Now().UTC()
+	defer func() {
+		logger.Debug("VerifyPresentation", log.WithDuration(time.Since(startTime)))
```
bstasyszyn, line 72: Do we need a metric here? (And in the other spots below.)

bstasyszyn (approved):

pkg/service/verifypresentation/verifypresentation_service.go
skynet2: @bstasyszyn ideally that should be spans in our tracing system, but as its not yet ready, this is the only way we can have at least some insights on performance, once tracing system will be in place, I will remove that and add spans

fqutishat (approved):

PR #1202, opened 2023-03-20
feat: tracing support for oidc4vci

aholovko:
PII data, specifically "claims" in pre-auth flow, undergoes sanitization to ensure privacy and security.

To monitor interactions with MongoDB, the database client is instrumented, providing detailed tracing capabilities on the level of queries.

For services, custom wrappers are used to keep business logic clean.

Lastly, controllers are manually instrumented with dedicated spans with request recording.

The global error handler has been updated to keep tracking between errors and traces. This is achieved by logging trace_id, which correlates error with the trace context where error occurs.


cmd/vc-rest/startcmd/start.go
```diff
@@ -458,6 +481,8 @@ func buildEchoHandler(
 		return nil, fmt.Errorf("failed to instantiate new oidc4ci service: %w", err)
 	}
 
+	oidc4ciService = oidc4citracing.Wrap(oidc4ciService, conf.Tracer)
```
bstasyszyn, line 484: I think this should be wrapped only if tracing is enabled, otherwise we're doing extra work (such as redacting) for no reason. (Same with other wrappers.)

bstasyszyn (approved):

skynet2 (approved):

PR #1227, opened 2023-03-31
feat: Add CSL Manager

sandrask:
Add CSL Manager


cmd/vc-rest/startcmd/start.go
```diff
@@ -388,6 +389,21 @@ func buildEchoHandler(
 		return nil, err
 	}
 
+	vcCrypto := crypto.New(conf.VDR, documentLoader)
+
+	vcStatusStore := vcstatusstore.NewStore(mongodbClient)
+
+	cslManager, err := cslmanager.New(
```
bstasyszyn, line 396: I think you need to move this to the credentialstatus component so that it may be overridden.

fqutishat: @bstasyszyn it's look correct

fqutishat (approved):

cmd/vc-rest/startcmd/start.go
bstasyszyn: > @bstasyszyn it's look correct

The CSL manager should be overridable in the credentialstatus component, so it should be moved to credentialstatus.

sandrask: Done

PR #1235, opened 2023-04-05
feat: validate issuer profile on revoke

mishasizov-SK:


test/bdd/pkg/v1/vc/credential.go
```diff
@@ -158,6 +158,14 @@ func (e *Steps) verifyRevokedVC(profileName, organizationName string) error {
 	return nil
 }
 
+func (e *Steps) revokeVCWithError(profileName, organizationName string) error {
+	if err := e.revokeVC(profileName, organizationName); err == nil {
+		return fmt.Errorf("error expected, but got nil")
```
fqutishat, line 163: can you check if we return correct error

fqutishat: credential status list does not belong to the current profile

fqutishat (approved):

PR #1257, opened 2023-04-14
feat: store temp data in redis

mishasizov-SK:


cmd/vc-rest/startcmd/params.go
```diff
@@ -221,11 +221,27 @@ const (
 	claimDataTTLFlagUsage = "Claim data TTL in OIDC4VC pre-auth code flow. Defaults to 3600s. " +
 		commonEnvVarUsageText + hostURLExternalEnvKey
 
+	claimDataStoreFlagName  = "claim-data-store"
```
fqutishat, line 224: we need to use one param for that

fqutishat (approved):

PR #1275, opened 2023-04-22
fix: removed additional http client params to unblock webhook

birtony:
Signed-off-by: Anton Biriukov <anton.biriukov@gendigital.com>


fqutishat (changes requested):
Why this change needed

birtony:
> Why this change needed

These additional parameters break demo issuer/verifier webhooks @fqutishat

fqutishat (approved):

PR #1283, opened 2023-04-26
feat: Add dynamic query capability for presentation definition

sandrask:
Presentation definition filters parameter defines which claim fields to retain in the definition.

Initial PR will implements fields filter. Input descriptor and retain filters will be implemented in the future.

Fields filter is defined as a matcher against the id property of each field entry. Rules:
- exact string match
- empty string rule - including “” will match all field entries that have no id
- regular expression match when you encounter * (can be used for prefix, suffix or grouping)


bstasyszyn (approved):

fqutishat (approved):

pkg/restapi/v1/verifier/controller.go
```diff
@@ -350,6 +361,57 @@ func (c *Controller) initiateOidcInteraction(
...
+			exp := strings.ReplaceAll(id, wildcard, "."+wildcard)
+			r, err := regexp.Compile(exp)
+			if err != nil {
+				return false, fmt.Errorf("failed to compile regex=%s. %w", exp, err)
+			}
+
+			if r.MatchString(target) {
+				return true, nil
+			}
+		}
```
sudeshrshetty, line 409: I think, if id has any special characters we may need to escape it.

sudeshrshetty: otherwise regex compile might break

sandrask: I thought it was supposed to break it - I have a test that breaks it. Do we support special chars in id?

PR #1313, opened 2023-05-09
feat: send verifier claim keys to Jaeger

aholovko:


fqutishat (changes requested):
we are changing code let's not merge it now

fqutishat (changes requested):
we are changing code let's not merge it now

fqutishat (approved):

PR #1326, opened 2023-05-26
feat: dynamic client registration protocol (RFC7591) - initial support

aholovko:
This PR:
* replaces OAuth2Client interface in OIDC4CI controller with [golang.org/x/oauth2](http://golang.org/x/oauth2) package
* refactors fosite `insert client` functionality
* updates OpenAPI spec and adds initial handler and service implementations for dynamic client registration


pkg/restapi/v1/oidc4ci/controller.go
```diff
@@ -655,6 +700,45 @@ func (c *Controller) oidcPreAuthorizedCode(
 	return &validateResponse, nil
 }
 
+// OidcRegisterDynamicClient handles OIDC Dynamic Client Registration (POST /oidc/register).
+func (c *Controller) OidcRegisterDynamicClient(e echo.Context) error {
```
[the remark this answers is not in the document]
aholovko: that name is auto-generated, I agree that it would be nice to have the proper acronym here but this is what we have :(, btw it's consistent across controller (i.e. there are `OidcPushedAuthorizationRequest`, `OidcAuthorize`, etc,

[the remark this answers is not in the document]
DRK3: Oh OK, thanks for letting me know!

DRK3 (approved):

pkg/service/clientmanager/clientmanager_service.go
```diff
@@ -0,0 +1,87 @@
...
+
+// NewService creates a new Service instance.
+func NewService(store store) *Service {
+	return &Service{
+		store: store,
+	}
+}
+
+// CreateClientRequest is a request to create an OAuth2 client.
+type CreateClientRequest struct {
```
DRK3, line 38: `RegisterClientRequest` would be a slightly more accurate name as the spec refers to this process as "registering" a client: https://datatracker.ietf.org/doc/html/rfc7591#section-3.1

pkg/restapi/v1/oidc4ci/openapi.gen.go
```diff
@@ -35,6 +35,66 @@ type AccessTokenResponse struct {
...
+	Scope *string `json:"scope,omitempty"`
+
+	// Requested client authentication method for the token endpoint. Supported values: none, client_secret_post, client_secret_basic. None is used for public clients (native apps, mobile apps) which can not have secrets. Default: client_secret_basic.
+	TokenEndpointAuthMethod *string `json:"token_endpoint_auth_method,omitempty"`
+}
+
+// Response with registered metadata for the client.
+type ClientRegistrationResponse struct {
+	// Client identifier.
+	ClientId string `json:"client_id"`
```
DRK3, line 65: I noticed that a lot of these generated names also don't do abbreviations correctly (e.g. it shoes `Id` instead of `ID`). Is this a limitation of the automatic generation?

fqutishat (approved):

pkg/service/clientmanager/clientmanager_service.go
aholovko: CreateClientRequest is the type from service layer, while ClientRegistrationRequest is auto-generated by tool. They belong to different layers (service vs. presentation) and help with separation of concerns.

pkg/restapi/v1/oidc4ci/openapi.gen.go
aholovko: [oapi-codegen](https://github.com/deepmap/oapi-codegen/) that is used in vcs allows customizing generated names with `x-go-name` custom tag, but currently we rely on default behavior. We might reconsider that in future :-)

As for types - tool uses pointers to differentiate absent of value from empty one - this is crucial for validation.

DRK3: OK, yeah I suppose customizing generate names would be a bit better, but certainly not a big deal. For the types, I understanding having `*string`, but I also see `*[]string` for `redirectUris`, which is unnecessary as slices/arrays are already nillable.

DRK3: (but, again, not really important, just something to look out for)

PR #1331, opened 2023-06-06
feat: issuer API for registering oauth clients

aholovko:


pkg/service/clientmanager/clientmanager_service.go
```diff
@@ -34,8 +34,8 @@ func NewService(store store) *Service {
 	}
 }
 
-// CreateClientRequest is a request to create an OAuth2 client.
-type CreateClientRequest struct {
+// ClientMetadata contains the metadata for an OAuth2 client.
+type ClientMetadata struct {
 	Name                    string   `json:"client_name"`
```
sudeshrshetty, line 39: Are we supporting only these 3 client metadata in registration request? how about additional params like logo, policy, jwks etc

sudeshrshetty (commented):
Are you adding public endpoint in follow up PR?

pkg/service/clientmanager/clientmanager_service.go
aholovko: public endpoint + service was added in previous PR

the current PR adds layer of private API (issuer), so we end up with the following flow: public endpoint -> private endpoint (on issuer) -> service

additional metadata is added in upcoming PR

fqutishat (approved):

PR #1341, opened 2023-06-19
feat: pre-authorized_grant_anonymous_access_supported

mishasizov-SK:


skynet2:
hi @fqutishat  @sudeshrshetty  fyi this will introduce a **breaking change** for current pre-authorize profiles. 
as currently we don't have authorized_grant_anonymous_access_supported in profiles, it will be false by default, and wallet-cli also don't send this value for now. 

cc @mishasizov-SK @aholovko

mishasizov-SK:
@skynet2 correct. As we aligned with @sudeshrshetty, I'll add this field to each profile and set it to `true`.

fqutishat (approved):

PR #1346, opened 2023-06-21
fix: Incorrect types for some client registration values

DRK3:


test/bdd/pkg/v1/oidc4vc/models.go
```diff
@@ -48,10 +48,10 @@ type clientRegistrationRequest struct {
 
 type clientRegistrationResponse struct {
 	ClientId                string                  `json:"client_id"`
-	ClientIdIssuedAt        string                  `json:"client_id_issued_at"`
+	ClientIdIssuedAt        *int                    `json:"client_id_issued_at,omitempty"`
```
aholovko, line 51: should be int

aholovko (changes requested):

test/bdd/pkg/v1/oidc4vc/models.go
DRK3: Fixed

aholovko (approved):

rolsonquadras (approved):

PR #1352, opened 2023-06-28
fix: "aud" claim check

DRK3:
To conform to the OpenID4CI spec, the "aud" claim check that happens on the credential endpoint now checks for the entire issuer URI instead of only the base URI.


pkg/restapi/v1/oidc4ci/controller.go
```diff
@@ -571,6 +584,10 @@ func (c *Controller) OidcCredential(e echo.Context) error {
 		var interactionErr *interactionError
 
 		if errors.As(parsedErr, &interactionErr) {
+			if interactionErr.Code == resterr.InvalidOrMissingProofOIDCErr {
+				return resterr.NewOIDCError(string(resterr.InvalidOrMissingProofOIDCErr), errors.New(interactionErr.Message))
+			}
```
mishasizov-SK, line 589: can it be done as another case in switch below ?

mishasizov-SK (changes requested):

aholovko (approved):

pkg/restapi/v1/oidc4ci/controller.go
DRK3: Yes, good catch! Just updated the PR

PR #1354, opened 2023-06-30
feat: update client registration to use public endpoint with profile ID and version

aholovko:
This PR updates dynamic client registration to use `/oidc/{profileID}/{profileVersion}/register` public endpoint with profile ID and profile version as part of URL. This allows to get rid of corresponding private issuer's endpoint and `issuer_state` parameter used for correlation purpose.

If dynamic client registration is supported by the profile, the value for registration endpoint (`registration_endpoint`) can be resolved from issuer's `/.well-known/openid-configuration`.


docs/v1/openapi.yaml
```diff
@@ -80,6 +80,44 @@ paths:
       description: Returns openid-config.
       tags:
         - issuer
+  '/issuer/{profileID}/{profileVersion}/oauth2/register':
```
sudeshrshetty, line 83: it will still be public endpoint, not private.

sudeshrshetty (changes requested):

sudeshrshetty (approved):

PR #1375, opened 2023-08-01
feat: oauth 2.0 client id scheme

aholovko:


sudeshrshetty (approved):

cmd/vc-rest/startcmd/start.go
```diff
@@ -722,15 +723,24 @@ func buildEchoHandler(
...
+		HTTPClient:       getHTTPClient(metricsProvider.ClientDiscoverableClientIDScheme),
+		ProfileService:   issuerProfileSvc,
+		TransactionStore: oidc4ciTransactionStore,
+	})
+
 	oidc4civ1.RegisterHandlers(e, oidc4civ1.NewController(&oidc4civ1.Config{
 		OAuth2Provider:          oauthProvider,
 		StateStore:              oidc4ciStateStore,
 		IssuerInteractionClient: issuerInteractionClient,
+		ProfileService:          issuerProfileSvc,
```
sudeshrshetty, line 737: why we need issuer profile service now for client ID scheme?

aholovko: Issuer profile is used for dynamic client registration to check if profile supports it (and it makes sense to do that as soon as possible and on that level as `/oidc/register` endpoint accepts profile ID and version). In the previous version of code that job was done by client manager service but now that manager is used for both dynamic client registration and discoverable client id scheme so check was moved to a level above.

fqutishat (approved):

PR #1410, opened 2023-09-08
chore: remove last afgo and orb dependencies

Moopli:


rolsonquadras (approved):

Makefile
```diff
@@ -180,13 +180,7 @@ clean:
 
 .PHONY: update-aries
 update-aries:
```
DRK3, line 182: Can this target just be removed completely?

component/wallet-cli/internal/storage/leveldb/leveldb.go
```diff
@@ -0,0 +1,661 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
```
DRK3, line 2: Copyrights should be updated

Makefile
Moopli: Not sure if it's used elsewhere, so I've left it with a deprecation notice for now.

component/wallet-cli/internal/storage/leveldb/leveldb.go
Moopli: Done.

DRK3 (approved):

PR #1439, opened 2023-09-27
refactor: update kms-go&vc-go, use kms-crypto wrappers

Moopli:


fqutishat (changes requested):
can't merge now until we run stress test

rolsonquadras (approved):

skynet2 (approved):

PR #1582, opened 2024-01-31
fix: use issuer DID instead of URL when validating issuer

aholovko:


aholovko:
@fqutishat @bstasyszyn Wallet CLI (or other wallet) needs to get issuer DID from signed JWT returned for `/.well-known/openid-credential-issuer` request, i.e. issuer profile should use JWT to communicate its OpenID Issuer Configuration if it wants to support attestation functionality.

bstasyszyn (approved):

PR #1724, opened 2024-05-21
feat: add extra claims in vp

skynet2:


fqutishat (approved):

pkg/service/oidc4vp/oidc4vp_service.go
```diff
@@ -53,6 +53,12 @@ const (
 	customScopeProperty  = "_scope"
 )
 
+const (
+	additionalClaimFieldName        = "name"
```
sudeshrshetty, line 57: Name & Description are not additional, they are part of W3C VC Data model

PR #1738, opened 2024-07-01
feat: attachments support

skynet2:


fqutishat (approved):

pkg/service/oidc4vp/attachments.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"github.com/samber/lo"
+	"github.com/trustbloc/vc-go/util/maphelpers"
+	"github.com/trustbloc/vc-go/verifiable"
+)
+
+const (
+	AttachmentTypeRemote   = "RemoteAttachment"
+	AttachmentTypeEmbedded = "EmbeddedAttachment"
+	AttachmentDataField    = "uri"
```
sudeshrshetty, line 28: how about hash & hash-alg?

sudeshrshetty (approved):

pkg/service/oidc4vp/attachments.go
skynet2: yep, i will add checks for hash in next pr, thanks

PR #1766, opened 2024-09-30
feat: kms per profile & bdd for eddsa-rdfc-2022

skynet2:


pkg/kms/arieskms.go
```diff
@@ -188,13 +191,23 @@ func (km *KeyManager) NewVCSigner(
 }
 
 func createLocalSecretLock(keyPath string) (secretlock.Service, error) {
-	if keyPath == "" {
-		return nil, fmt.Errorf("no key defined for local secret lock")
+	var primaryKeyReader io.Reader
+	var err error
+	if os.Getenv("VCS_LOCAL_KMS_MASTER_KEY") != "" {
```
fqutishat, line 196: can you add it in parameter file in cmd

fqutishat (approved):

